PR #10, opened 2026-08-23
[systems-test] Fix the camera capture check and clean up the output

cprussin:
## The bug

`Checking if all cameras are capturing...` always failed.

Frigate 0.17's `/stats` route is guarded by `allow_any_authenticated()`, which rejects any request without a `remote-user` header — that header is only set by nginx's `auth_request` subrequest. Hitting the api upstream at `127.0.0.1:5001` directly bypasses nginx entirely, so it always answered 401. With `curl -sf` the body was suppressed too, so the failure printed an empty reason.

The nixpkgs frigate module already listens on `127.0.0.1:5000` for exactly this ([nixpkgs#370349](https://github.com/NixOS/nixpkgs/issues/370349)), and frigate's `/auth` endpoint stamps requests arriving on that port as `remote-user: anonymous` / `remote-role: admin`. So:

- Read stats from `http://127.0.0.1:5000/api/stats` instead of `http://127.0.0.1:5001/stats`.
- Swap `-f` for `--fail-with-body -S` so a future failure actually reports why.

## Output cleanup

Failures were hard to follow — the failure text was flush left and blended into the checks that followed it, and a long run gave no recap.

- Statuses are padded to a fixed column so they line up.
- Failure output is indented behind a red rule, so it reads as belonging to the check above it.
- Every failed check is listed again in the summary, with a `N of M` count.
- Colors only when stdout is a terminal.
- Descriptions drop their leading `if` so they read as noun phrases both inline and in the summary.

```
Checking crux is up on wireguard ipv6 .......................... ok
Checking all cameras are capturing ............................. FAILED
  │ Cameras not capturing: driveway, side
Checking immich is running ..................................... ok
Checking crux's zpools are healthy ............................. FAILED
  │ pool: tank
  │   state: DEGRADED
  │   scan: none requested

2 of 4 checks failed! ❌
  ✘ all cameras are capturing
  ✘ crux's zpools are healthy
```

## Testing

The generated shell passes `bash -n`, and the reporting scaffolding was exercised end to end with stubbed checks (pass, single-line failure, multi-line failure, silent failure) in both tty and piped modes. The frigate change itself needs a run against crux to confirm.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QBzTPt8seBNfjisXg5aTDM)_

### DIFF
--- a/systems-test.nix
+++ b/systems-test.nix
@@ -26,7 +26,25 @@ in
     bash="${pkgs.bash}/bin/bash"
     xmllint="${pkgs.libxml2}/bin/xmllint"
 
-    FAILURES=0
+    if [ -t 1 ]; then
+        bold=$'\e[1m'
+        dim=$'\e[2m'
+        red=$'\e[31m'
+        green=$'\e[32m'
+        reset=$'\e[0m'
+    else
+        bold=""
+        dim=""
+        red=""
+        green=""
+        reset=""
+    fi
+
+    # Column that check statuses are padded out to, so they line up.
+    STATUS_COLUMN=72
+
+    TOTAL=0
+    FAILURES=()
 
     # Run a check, printing its output only when it fails.  A check is any
     # command that exits nonzero and explains itself on stdout or stderr when
@@ -34,16 +52,30 @@ in
     check() {
         local description="$1"
         shift
-        echo -n "Checking $description...  "
+        TOTAL=$((TOTAL + 1))
+
+        # Pad the description out with dots so that the status column lines up
+        # and a failure is easy to spot while scanning down the output.
+        local label="Checking $description "
+        local width=$((STATUS_COLUMN - ''${#label}))
+        local dots=""
+        if [ "$width" -gt 0 ]; then
+            dots=$(printf "%''${width}s" "" | tr " " ".")
+        fi
+        printf "%s%s%s%s " "$label" "$dim" "$dots" "$reset"
+
         local output
         if output=$("$@" 2>&1); then
-            echo "✅"
+            printf "%sok%s\n" "$green" "$reset"
         else
-            echo "❌"
-            if [ -n "$output" ]; then
-                echo "$output"
-            fi
-            FAILURES=$((FAILURES + 1))
+            printf "%s%sFAILED%s\n" "$bold" "$red" "$reset"
+            FAILURES+=("$description")
+            # Indent the failure behind a rule so it reads as belonging to the
+            # check above it rather than to the checks that follow.
+            local line
+            while IFS= read -r line; do
+                printf "  %s│%s %s\n" "$red" "$reset" "$line"
+            done <<<"''${output:-(the check failed without explaining itself)}"
         fi
     }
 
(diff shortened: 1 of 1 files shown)


cprussin (commented):
LGTM — no issues found. (Submitting as a comment rather than an approval because GitHub won't let this account approve its own PR.)

I verified the frigate claim independently and exercised the generated shell rather than trusting the description.

**Frigate fix — confirmed correct** against the pinned nixpkgs (`d407951`) and frigate `v0.17.1`:

- `nixos/modules/services/video/frigate.nix` puts `listen 127.0.0.1:5000;` in `virtualHosts."${cfg.hostname}".extraConfig`, and in the nginx module `extraConfig` is rendered into the **main** server block (`forceSSL`'s `return 301` lives in a separate `redirectListen` block on :80 bound only to `listenAddresses`). So `http://127.0.0.1:5000` is plain HTTP against the real vhost — `forceSSL = true` in `nvr.nix` does not redirect it, and nothing else in this repo binds :5000.
- The vhost has `location /api/stats` with `nginxAuthRequest` + `rewrite ^/api(/.*)$ $1 break; proxy_pass http://frigate-api;` → hits frigate's `/stats`, so `/api/stats` (not `/stats`) is the right path through :5000.
- `location /auth` sets `proxy_set_header X-Server-Port $server_port;`, and `frigate/api/auth.py:577` does `if int(request.headers.get("x-server-port", default=0)) == 5000:` → returns `remote-user: anonymous` / `remote-role: admin`. `/stats` is `Depends(allow_any_authenticated())` (`app.py:85`) and is in `require_admin_by_default`'s `EXEMPT_PATHS`, so it passes. Direct 5001 has no `remote-user` → 401, exactly as the comment says.
- `stats["cameras"][name]["camera_fps"]` still exists (`frigate/stats/util.py:281`), so the existing jq filter is valid against the real payload.
- `curl` is in NixOS `requiredPackages` (`nixos/modules/config/system-path.nix`), so it's on crux's PATH under `ssh` BatchMode; `--fail-with-body` + `-sS` correctly puts both the body and curl's error text into `$stats` via the `2>&1` on the local `on_crux` call.

**Nix string** — tokenized the `''`-string by hand (no `nix` binary available): every `''` in the body is `''${`, which continues the string; the literal terminates exactly at the final `''` with only a newline after. Interpolations are all valid attribute paths. `\e` / `\n` stay literal backslashes as intended for `$'...'` and `printf`.

**Shell** — extracted the generated script, `bash -n` clean, and ran the reporting scaffolding with stubbed checks in both tty and piped modes:
- Passing, silent-failure, multi-line-failure, output containing `%` and `\`, and output starting with `-e`/`-n` all render correctly — `printf "%s"` avoids the `echo` pitfalls the old code had.
- `${output:-...}` placeholder fires only on genuinely empty output; the herestring guarantees the last line is read.
- Non-tty run emits no escapes; `${#FAILURES[@]}` / `TOTAL` and the `exit 1` are right.
- `width <= 0` is guarded; the longest real label is `Checking the certificate for home-assistant.internal.prussin.net ` at 65 chars, comfortably under `STATUS_COLUMN=72`.
- `while ... done <<<` runs in the current shell, so no lost state; nothing else in the repo parses this script's output (only `cli.nix` references it).

One behavioral consequence worth knowing, not a defect: the camera check now goes through nginx, so an nginx outage on crux fails both `eyes is running` and `all cameras are capturing` instead of just the former.

---

PR #12, opened 2026-08-24
[backup] Replace the scrub watch with a redrawing progress loop

cprussin:
`run-backup` stalled at `Scrubbing tank-backup...` printing `watch: getchar(): Success`.

**Cause:** procps-ng's `watch` reads stdin for interactive keys; backgrounded (`watch -n1 ... &`) its `getchar()` fails immediately, so it exits and nothing is ever shown. `zpool scrub -w` then ran to completion with no output.

**Fix:** drop `watch` and poll `zpool status` in the foreground:

- On a terminal: print the `scan:` block, then `ESC[<n>A ESC[J` back over it each second so it refreshes in place. Lines are `expand`ed and truncated to the terminal width first, so one logical line is always one row and the cursor-up count stays right on a narrow window. Width is read through a saved copy of stdout (`exec 3>&1`; `tput cols <&3`), since `tput` inside `$( )` otherwise measures stderr or stdin.
- Not a terminal: no progress output at all — the final status is the only thing printed, instead of a near-identical block every second for hours.
- A failing `zpool status` breaks the loop rather than tripping `set -e`, so the cleanup below (export, lock, snapshot destroy) still runs.
- On completion the last progress block is cleared and the full status (config + errors) prints once.

Verified against a stubbed `zpool`, rendered through pyte in a 45x30 pty: in-place redraw with no residue, earlier output preserved, correct width with stdin redirected from `/dev/null`, only the final status when piped, and cleanup reached with exit 0 when `zpool status` fails.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CDqamJFvqXMeXv3TmDqBev)_

### DIFF
--- a/config/machines/crux/backup.nix
+++ b/config/machines/crux/backup.nix
@@ -53,11 +53,33 @@
 
     echo
     echo -e "\e[1;37mScrubbing tank-backup...\e[0m"
-    watch -n1 zpool status tank-backup &
-    PID=$!
-    zpool scrub -w tank-backup
-    kill -INT $PID
-    zpool status tank-backup
+    zpool scrub tank-backup
+    exec 3>&1
+    DRAWN=0
+    clearProgress() {
+      if [ -t 1 ] && [ "$DRAWN" -gt 0 ]
+      then
+        printf '\033[%dA\033[J' "$DRAWN" || true
+        DRAWN=0
+      fi
+    }
+    while true
+    do
+      NEXT=$(zpool status tank-backup 2>&1) || { clearProgress; echo "$NEXT" >&2; break; }
+      STATUS=$NEXT
+      clearProgress
+      echo "$STATUS" | grep -q 'scrub in progress' || break
+      if [ -t 1 ]
+      then
+        WIDTH=$(tput cols <&3 2>/dev/null || echo 80)
+        PROGRESS=$(echo "$STATUS" | sed -n '/scan:/,/^config:/{/^config:/!p}' | expand | cut -c "1-$WIDTH")
+        echo "$PROGRESS"
+        DRAWN=$(echo "$PROGRESS" | wc -l)
+      fi
+      sleep 1
+    done
+    exec 3>&-
+    echo "$STATUS"
 
     echo
     echo -e "\e[1;37mCleaning up...\e[0m"


config/machines/crux/backup.nix
```diff
@@ -53,11 +53,22 @@
...
+    do
+      STATUS=$(zpool status tank-backup)
+      if [ -t 1 ] && [ "$LINES" -gt 0 ]
+      then
+        printf '\033[%dA\033[J' "$LINES"
+      fi
+      echo "$STATUS" | grep -q 'scrub in progress' || break
+      PROGRESS=$(echo "$STATUS" | sed -n '/scan:/,/^config:/{/^config:/!p}')
+      echo "$PROGRESS"
+      LINES=$(echo "$PROGRESS" | wc -l)
```
cprussin: `LINES` counts *logical* lines, but `\033[NA` moves up *terminal rows*. If any progress line is longer than the terminal width, the terminal consumes more rows than `wc -l` reports, so the cursor-up under-counts, lands *inside* the previous block, and `\033[J` erases only from there down. The top `rows - LINES` rows of each block survive, and the residue accumulates one block per second for the whole scrub.

Not hypothetical on a narrow window/split pane — the middle scan line is already close to 80 columns once the leading tab is expanded, e.g.

```
	4.79T / 8.19T scanned at 1.20G/s, 1.15T / 8.19T issued at 295M/s
```

Cheapest fix is to truncate to the terminal width so one logical line is always one row:

```sh
WIDTH=$(tput cols 2>/dev/null || echo 80)
PROGRESS=$(echo "$STATUS" | sed -n '/scan:/,/^config:/{/^config:/!p}' | expand | cut -c "1-$WIDTH")
```

(`expand` first because `cut -c` counts characters, and the tab-indented continuation lines display as 8 columns.)

---

config/machines/crux/backup.nix
```diff
@@ -53,11 +53,22 @@
...
+    while true
+    do
+      STATUS=$(zpool status tank-backup)
+      if [ -t 1 ] && [ "$LINES" -gt 0 ]
+      then
+        printf '\033[%dA\033[J' "$LINES"
+      fi
+      echo "$STATUS" | grep -q 'scrub in progress' || break
+      PROGRESS=$(echo "$STATUS" | sed -n '/scan:/,/^config:/{/^config:/!p}')
+      echo "$PROGRESS"
```
cprussin: When stdout isn't a tty the redraw is correctly skipped — but the block is still printed every second. Piping `run-backup` to a log or `tee` during a multi-hour scrub emits tens of thousands of near-identical progress blocks, where the old `watch` version emitted essentially nothing useful.

Worth gating the whole progress print on `[ -t 1 ]` and letting the trailing `echo "$STATUS"` be the only output in the non-interactive case.

---

config/machines/crux/backup.nix
```diff
@@ -53,11 +53,22 @@
...
-    watch -n1 zpool status tank-backup &
-    PID=$!
-    zpool scrub -w tank-backup
-    kill -INT $PID
-    zpool status tank-backup
+    zpool scrub tank-backup
+    LINES=0
+    while true
+    do
+      STATUS=$(zpool status tank-backup)
```
cprussin: Minor, take it or leave it: under `set -e` a failing `zpool status` here aborts the script, skipping the whole cleanup block (`zpool export`, `systemctl stop unlock-…`, `zfs destroy`) and leaving the backup pool imported and the LUKS device unlocked. That's how the rest of this script already behaves (there's no `trap`), so it isn't a regression — but this used to be one `zpool scrub -w` call and is now one call per second for hours, so it's a lot more chances to trip.

`|| break` would fall through to the cleanup instead, at the cost of `echo "$STATUS"` printing empty.

---

cprussin (commented):
The core logic is sound — nice improvement over the `watch`/`kill -INT` dance. One real (cosmetic) bug and two minor notes inline; nothing blocking.

Things I specifically checked and found correct:

- **Nix escaping.** No `${` is introduced (`$(…)`, `$STATUS`, and the `sed` `{…}` block don't antiquote), so nothing needs `''${`. `\` isn't an escape character in `''…''`, so `\033[%dA\033[J` reaches the shell verbatim, and bash's `printf` renders `\033` as ESC (verified). `%d` is untouched by Nix.
- **`sed` extraction.** Verified against real-shaped output: `  scan:` is unanchored so it matches the two-space indent; the continuation lines are tab- (modern) or space- (older) indented and never match `^config:`; `config:` is always at column 0 and is correctly suppressed by `{/^config:/!p}`. Extra sections that ZFS prints between `scan:` and `config:` (`checkpoint:`, removal, raidz-expand) get swept in, which is harmless. Both degenerate cases stay self-consistent with `LINES`: no `scan:` → empty `PROGRESS` → `echo` prints one blank line and `wc -l` says 1; no `config:` → range runs to EOF and `LINES` still matches what was printed.
- **Termination.** Breaks on completed, `scrub canceled`, `scrub paused`, and on a resilver preempting the scrub — all of those drop the `scrub in progress` substring. `grep … || break` is `set -e`-safe, and `STATUS` is always assigned before it's read.
- **Redraw ordering.** Erasing *before* the `break` test means the final `echo "$STATUS"` cleanly replaces the progress block rather than appending under it. Nice.
- **No scrub-start race.** `zpool scrub` goes through `ZFS_IOC_POOL_SCAN` → `dsl_scan_setup_sync`, which is a synchronous sync task, so the scan state is committed before the command returns and the first poll sees `scrub in progress`. And if a scrub genuinely does finish before the first poll, breaking immediately and printing the final status is the right outcome anyway.

---

config/machines/crux/backup.nix
cprussin: Fixed as suggested — `expand | cut -c "1-$WIDTH"` with `WIDTH=$(tput cols 2>/dev/null || echo 80)`. Verified `tput cols` still reports the real width inside command substitution (133 in a 133-column pty), and at 45 columns every emitted progress line is exactly ≤45 chars with no tabs, so one logical line is one row.

---

config/machines/crux/backup.nix
cprussin: Done — the whole progress print is now inside `if [ -t 1 ]`, so a piped run emits only the final `echo "$STATUS"`.

---

config/machines/crux/backup.nix
cprussin: Taken. It's now `NEXT=$(zpool status tank-backup) || break; STATUS=$NEXT`, so a failed poll falls through to cleanup while `STATUS` keeps the last good output for the final print — only a failure on the very first poll prints empty. Verified: with a stubbed `zpool status` that fails, the loop breaks and the cleanup block is reached with exit 0.

---

config/machines/crux/backup.nix
```diff
@@ -53,11 +53,27 @@
...
+      then
+        printf '\033[%dA\033[J' "$LINES"
+      fi
+      echo "$STATUS" | grep -q 'scrub in progress' || break
+      if [ -t 1 ]
+      then
+        WIDTH=$(tput cols 2>/dev/null || echo 80)
+        PROGRESS=$(echo "$STATUS" | sed -n '/scan:/,/^config:/{/^config:/!p}' | expand | cut -c "1-$WIDTH")
+        echo "$PROGRESS"
+        LINES=$(echo "$PROGRESS" | wc -l)
```
cprussin: Blocking: `LINES` is a bash-special variable — the script never gets to keep the value `wc -l` computed.

`checkwinsize` is on by default since bash 5.0, **including non-interactive shells**, so bash re-binds `LINES` (and `COLUMNS`) to the terminal's row count after running commands whenever its tty ioctl succeeds — i.e. exactly in the `[ -t 1 ]` case where the redraw runs. So `printf '\033[%dA\033[J'` gets the *screen height*, not the block height: it jumps to the top of the visible screen and `\033[J` wipes everything on it, once per second, for the whole scrub. That destroys all the earlier output of the run (the "Mounting…", "Creating backup snapshot…", "Sending incremental data…" sections), which is worse than the residue this round was fixing.

Reproduced with the loop verbatim, a stubbed `zpool`, and a 45x30 pty (rendered through pyte):

```
$ ... emitted escape sequence
\x1b[30A\x1b[J        # 30 = terminal rows, not the 3 progress lines
```
Final screen — the line printed before the loop is gone:
```
|  pool: tank-backup|
| state: ONLINE|
...
```

Minimal proof of the clobber, non-interactive bash 5.2 under a pty:
```
immediately after assign: LINES=3 DRAWN=3
after running a command:  LINES=30 DRAWN=3
```

Fix is just the variable name (`DRAWN`, `ROWS`, anything non-special). Same simulation after renaming emits `\x1b[3A\x1b[J` and the preceding line survives:
```
|=== before ===|
|  pool: tank-backup|
...
```

---

config/machines/crux/backup.nix
```diff
@@ -53,11 +53,27 @@
...
+      NEXT=$(zpool status tank-backup) || break
+      STATUS=$NEXT
+      if [ -t 1 ] && [ "$LINES" -gt 0 ]
+      then
+        printf '\033[%dA\033[J' "$LINES"
+      fi
+      echo "$STATUS" | grep -q 'scrub in progress' || break
+      if [ -t 1 ]
+      then
+        WIDTH=$(tput cols 2>/dev/null || echo 80)
```
cprussin: Non-blocking, but the width here is not measured from the fd being drawn on.

ncurses `tput` picks its terminal fd as stdout → stderr → stdin (first one that's a tty). Inside `$( )` stdout is the pipe, and the `2>/dev/null` makes stderr `/dev/null` too — so the 133 verified in round one actually came from **stdin**. When stdin isn't the terminal, `tput` silently falls back to the terminfo default (80) even though stdout is a real, possibly narrower terminal, and the wrap residue is back. Measured in a 45-column pty:

```
stdin=/dev/null stderr=/dev/null : 80    <- wrong, terminal is 45
stdin=/dev/null stderr=tty       : 45
stdin=<&3 (saved real stdout)    : 45
```

So `sudo run-backup </dev/null` (or any wrapper that redirects stdin) on a sub-80-column terminal silently reintroduces the bug. Measuring the fd you're actually writing to fixes it:

```sh
exec 3>&1
...
WIDTH=$(tput cols <&3 2>/dev/null || echo 80)
```

---

cprussin (commented):
Round two. All three round-one fixes are in and the structure is right, but the redraw is still broken for a reason the truncation fix can't cover: `LINES` is a bash-special variable and bash overwrites it with the terminal height, so the cursor-up count is the screen height every iteration. Details inline — that one needs a change.

Checked and fine, no action needed:

- **`NEXT=$(...) || break; STATUS=$NEXT`** — correct. The assignment is the left side of an `||`, so `set -e` doesn't fire, and `break` lands in the cleanup path. An unset `STATUS` on a first-poll failure only makes the final `echo` print a blank line (no `set -u`), which is the documented trade-off.
- **`WIDTH` sanity** — no crash path found. ncurses only emits a positive integer or fails: `ws_col=0` from the ioctl and `COLUMNS` of `0`/`abc`/`-5` all fall back to 80; unset/bogus `TERM` exits non-zero into `|| echo 80`. And an empty `WIDTH` gives `cut -c "1-"`, which is valid (no truncation, exit 0). Worth knowing that the failure would have been nasty though — `cut -c "1-0"` exits 1, which under `set -e` aborts the script mid-scrub and skips the whole cleanup block, leaving the pool imported and the LUKS device unlocked.
- **PATH / bare command names** — safe. `pkgs.ncurses` (tput) and `pkgs.coreutils-full` (expand, cut) are both in NixOS's `requiredPackages`, so they resolve from `/run/current-system/sw/bin` exactly like the `zpool`/`zfs`/`grep`/`sed` this script already calls bare, and sudo preserves `PATH` and `TERM`. `tput` additionally has the `|| echo 80` fallback; `expand`/`cut` don't, but coreutils is never absent.
- **`[ -t 1 ]` gating vs. bookkeeping/ordering** — consistent. `LINES` can only become non-zero inside the tty branch, so the erase can't run in the non-tty case, and the erase-before-`break` ordering does clear the last block before the final status. Verified end-to-end with a stubbed `zpool`: piped, the run emits only the final status and reaches cleanup with exit 0.
- **Nix `''` escaping** — clean. Nothing new contains `${` or a `''` pair; `\033` is literal inside an indented string and reaches `printf` intact.

---

config/machines/crux/backup.nix
cprussin: Confirmed and fixed — renamed to `DRAWN`.

My round-one check missed this because I probed with a builtin (`true`), which doesn't trigger the reset; after any *external* command bash rebinds it. Non-interactive bash 5.2 in a 45x30 pty:

```
start:            LINES=3 DRAWN=3
after builtin:    LINES=3 DRAWN=3
after external:   LINES=30 DRAWN=3
after cmd subst:  LINES=30 DRAWN=3
```

Rendered the loop verbatim through pyte at 45x30: before the rename it emits `\x1b[30A` and the line printed ahead of the loop is gone; after, it emits `\x1b[3A` and that line survives, with only the final status left on screen.

---

config/machines/crux/backup.nix
cprussin: Right — took the suggested fix: `exec 3>&1` before the loop, `WIDTH=$(tput cols <&3 2>/dev/null || echo 80)`, `exec 3>&-` after. Reproduced your measurement in a 45-col pty (stdin `/dev/null`, stderr `/dev/null` → 80; via `<&3` → 45), and the full loop now renders correctly under pyte with the script's stdin redirected from `/dev/null`.

---

cprussin:
## Round-three review — commit `9ffba3d`

_(Posted as a PR comment rather than a formal review: the GraphQL review-write endpoint has been returning `API rate limit already exceeded` for the last ~40 minutes. Content is unchanged; nothing here is blocking.)_

**Verdict: no blocking defects.** The round-two fixes are correct, and apart from one cosmetic artifact on the error path (below) they introduced nothing new.

Everything below was run for real — bash 5.2 under actual ptys, rendered through `pyte`, against a stubbed `zpool`, using the loop body copied verbatim out of the Nix string.

### fd 3 handling — correct and self-contained
- Probed fd 3 from a genuine external child (`/bin/bash -c '[ -e /proc/self/fd/3 ]'`). Note that both `ls /proc/self/fd` and `$( )` report a **false positive** here from their own fds, which is easy to get wrong. Result: open for loop children only, closed before `zpool export` / `systemctl stop` / `zfs destroy`.
- `tput cols <&3` works even when fd 1 is a **write-only** tty fd (`open(ttyname, O_WRONLY)` then `exec 3>&1`): returned the real 45, not the 80 fallback. So `run-backup > /dev/pts/N` still measures the right terminal.
- Not reaching `exec 3>&-` is harmless — fd 3 is just a dup of stdout and dies with the process. `exec 3>&-` on an unopened fd is also rc=0 under `set -e`.
- With stdin `/dev/null` (the `sudo` shape) the loop renders correctly.

### Variable names — all five clear
```
DRAWN      = [SENTINEL]      LINES   = [30]   <- still clobbered; mechanism confirmed live
WIDTH      = [SENTINEL]      COLUMNS = [45]   <- ditto
NEXT       = [SENTINEL]
STATUS     = [SENTINEL]
PROGRESS   = [SENTINEL]
```
measured after external commands and command substitution, with `shopt -s checkwinsize` active. None appear in `env -i bash --norc -c compgen -v`, and none collide with the script's own `TODAY` / `LAST_SNAP_DATE`.

### Redraw correctness
- 45x30 and 60x24: emits `\x1b[3A`, the line printed before the loop survives, no residue, cleanup reached, exit 0.
- Variable block height (3/2/3/2 lines across iterations): emitted `[3, 2, 3, 2, 3]` — the erase always uses the height of the block actually drawn, so a scan section that grows or shrinks mid-scrub is handled.
- A progress line **exactly** `$WIDTH` chars wide still occupies one row (deferred wrap) — no off-by-one from `cut -c "1-$WIDTH"`.
- Pathological 3-row and 2-row terminals: clean, no residue.

### `set -e` / cleanup
Every command in the loop is either in an `if` condition or guarded by `||`. The hazard I went looking for was `cut -c "1-$WIDTH"` aborting on a bad width (`1-0` is a hard error, rc=1) — not reachable: `tput cols` returns either the real width or 80 for `TERM` in {xterm-256color, dumb, linux, vt100, unset, bogus}, and returns 80 (not 0) even on a pty whose `TIOCGWINSZ` is 0x0. An empty `$WIDTH` would give `1-`, which is also valid ("to end of line").

### Terminal state on abort
SIGINT mid-loop: the loop never emits a single `\x1b[?Nh/l` — no alternate screen, no cursor hiding — so the terminal is left clean with just the last progress block visible. Strictly better than the old version, which orphaned a `watch` child.

### Non-tty stdout
Zero escape sequences, one final status block, rc=0.

### Nix escaping
The only `''` in the string are its delimiters, there is no `${` anywhere in the added lines, and `<&3` / `3>&1` / `3>&-` contain nothing Nix-special. `\033` passes through literally in an indented string, exactly like the pre-existing `\e[1;37m`.

---

### Non-blocking, introduced by round two's `|| break`
The `break` fires *before* the erase, so on a failed poll the stale progress block is never cleared. Measured with a `zpool status` stubbed to fail on the third poll, 60x24:

```
|=== ANCHOR LINE MUST SURVIVE ===|
|Scrubbing tank-backup...|
|  scan: scrub in progress since now|      <- stale block, never erased
|        11.04% done|                      <-
|cannot open 'tank-backup': no such pool|
|  pool: tank-backup|                      <- final echo "$STATUS": same content again
|  scan: scrub in progress since now|
|        11.04% done|
```

So the user sees the last progress block twice with the error wedged between. Purely cosmetic, only on the failure path, and cleanup still runs (exit 0) — which was the point of the change. Fixable by moving the erase above the `|| break` if you care; fine to leave.

### Resize mid-scrub — checked, nothing to fix
`tput cols <&3` re-measures every iteration, so truncation tracks the new width immediately (verified: resized 100→45 mid-loop; subsequent lines truncate at 45 and the bookkeeping stays consistent). The only residual case is a terminal that *reflows* scrollback on resize (VTE, kitty, alacritty — not xterm or tmux): a block drawn at 100 cols becomes ~6 rows after narrowing to 45, the next `\x1b[3A\x1b[J` under-counts, and up to block-height stale rows get stranded once. Steady state resumes on the very next iteration, and widening cannot produce it at all since lines are pre-truncated and never wrap. I could not reproduce this empirically — `pyte` does not reflow — so that last case is analysis rather than measurement. It is bounded, one-time per resize, and inherent to redrawing in the normal buffer; avoiding it is what the alternate screen bought `watch`, which is the thing being removed.

---

cprussin:
Round three addressed in `6db613f`.

**The failure-path artifact:** fixed rather than left. Erasing before `break` alone wasn't enough — the error goes to the terminal *between* the block and the erase, so `ESC[3A` lands one row inside the block and the first `scan:` line survives (reproduced exactly that). So the poll now captures stderr and replays it after the erase:

```sh
clearProgress() {
  if [ -t 1 ] && [ "$DRAWN" -gt 0 ]
  then
    printf '\033[%dA\033[J' "$DRAWN"
    DRAWN=0
  fi
}
...
NEXT=$(zpool status tank-backup 2>&1) || { clearProgress; echo "$NEXT" >&2; break; }
```

Measured at 60x24 with `zpool status` stubbed to fail on the third poll — block fully cleared, error shown once, last-good status printed once, no duplication, cleanup reached, exit 0:

```
|=== ANCHOR LINE MUST SURVIVE ===|
|cannot open 'tank-backup': no such pool|
|  pool: tank-backup|
| state: ONLINE|
...
```

First-poll failure is also clean (error, blank status line, cleanup). Normal completion, piped output, and 45x30 with stdin `/dev/null` all unchanged from your round-three results.

**Reflow-on-resize:** agreed, leaving it. Bounded, self-correcting on the next iteration, and avoiding it means the alternate screen — which is what's being removed.

---

cprussin:
_(Posted as a PR comment rather than a formal review: the GraphQL review-write endpoint is still returning `API rate limit already exceeded`, as it was for round three. Content is unchanged; nothing here is blocking.)_

## Round-four review — commit `6db613f`

**Verdict: no blocking defects.** Scoped to the only delta since round three (`clearProgress` + `2>&1` on the poll). `git diff 9ffba3d..6db613f` is exactly the four hunks you described and nothing else, so everything settled in rounds one to three still stands.

Everything below was measured — bash 5.2 under real ptys rendered through `pyte`, the loop body copied verbatim out of the Nix string, against a stubbed `zpool`.

### Folding stderr into `NEXT` — safe on the success path, and it closes a latent bug

Checked a `zpool status` that exits 0 while writing to stderr, both single- and multi-line, and with the warning forced to land *inside* the `scan:`/`config:` range (worst-case flush ordering):

- `grep -q 'scrub in progress'` — unaffected.
- `sed -n '/scan:/,/^config:/{/^config:/!p}'` — a warning outside the range is not extracted; one forced *inside* the range is drawn as an extra row, but `DRAWN=$(echo "$PROGRESS" | wc -l)` counts the same text that was echoed, so the erase stays exact. No stale rows in either case.
- Final `echo "$STATUS"` — the warning is printed once, at the end.

It also **fixes** something round three had. With the old `NEXT=$(zpool status tank-backup)`, a successful poll's stderr went straight to the terminal between the draw and the erase, so the cursor-up landed inside the block. Reproduced on the round-three tree, 60x20:

```
|WARNING: zfs-kmod version mismatch|
|  scan: scrub in progress since Mon Aug 24 11:59:00 2026|   <- stale
|  scan: scrub in progress since Mon Aug 24 11:59:00 2026|   <- stale
|  scan: scrub in progress since Mon Aug 24 11:59:00 2026|   <- stale
```

Same scenario on `6db613f` is clean. So `2>&1` isn't only about the failure path.

Two behaviour changes worth knowing about, neither a defect:
- stderr from a *successful* poll is now swallowed unless it happens on the final poll. A transient vdev warning mid-scrub is no longer shown. Arguably right — it would otherwise be reprinted every second.
- when it does survive, it comes out of `echo "$STATUS"` on **stdout**. Under `run-backup >log 2>errlog` a `zpool status` warning now lands in `log` rather than `errlog`.

### `clearProgress` under `set -e` — correct in both positions

- Branch not taken: the `if` compound returns 0, so the function returns 0. Measured 0 with `DRAWN` unset and with `DRAWN=notanumber` too (`[: integer expression expected` is rc 2, but it's the `if` condition, so it's swallowed and the compound still yields 0).
- Branch taken: last command is the `DRAWN=0` assignment, rc 0.
- This matters, because errexit is **not** suppressed inside the `||` RHS group — that group is "the command following the final `||`". Control test: a function returning 3 in that position kills the script with rc 3 before the group's later commands run. So the failure path reaches `echo "$NEXT" >&2` and `break` only because `clearProgress` is unconditionally 0. It is, in every state I could put it in.

### `DRAWN` bookkeeping — no divergence found

- Both call sites run in the current shell (the `||` RHS group is not a subshell — confirmed `DRAWN` mutates through it), so the reset lands.
- The added `DRAWN=0` is what makes the function idempotent, which is what makes it safe to call on a path where nothing was drawn (first-poll failure) and on a path where it may be called twice.
- `DRAWN` is only ever set nonzero inside `if [ -t 1 ]`, and the erase is gated on the same test, so the non-tty path can't emit escapes. Verified: `cat -A` on piped output shows zero escape sequences.
- Variable block height across iterations (3/4/3/4 rows) tracks correctly; `PROGRESS` empty (no `scan:` line) gives one blank row and `DRAWN=1`, cleared correctly next iteration.

### Name collision — none

`clearProgress` appears nowhere else in the repo, and no such command exists. An exported `BASH_FUNC_clearProgress%%` in the environment cannot shadow it — the in-script definition wins (verified), and `sudo` strips those anyway.

### Nix escaping — clean

No `''` and no `${` anywhere in the added lines; the only `${` in the whole `runBackup` string are the four pre-existing `config.*` interpolations. `\033` is a literal backslash-0-3-3 in the file (`od -c` confirms), passed through untouched by the indented string and interpreted by `printf`. `2>&1`, `>&2` and the `{ ...; }` group contain nothing Nix-special.

### Empirical results

Failure at poll 3, 60x20 — no stale rows, error once, last-good status once, cleanup reached, exit 0:

```
|=== ANCHOR LINE MUST SURVIVE ===|
|Scrubbing tank-backup...|
|cannot open 'tank-backup': no such pool|
|  pool: tank-backup|
| state: ONLINE|
...
|Cleaning up...|
|CLEANUP REACHED|
```

First poll fails (`STATUS` never assigned) — no `set -u`, so `echo "$STATUS"` is just an empty line; error shown once, cleanup reached, exit 0. The only artifact is that one stray blank line, which predates this commit (it came in with round two's `|| break`). Not worth a change.

Non-tty failure path: error on stderr, no escapes on stdout, exit 0. Normal completion and the piped case are byte-identical to round three.

---

### Non-blocking: `printf` failure inside `clearProgress` can now abort the failure path

Errexit is live inside the function, so if the `printf` fails the script dies there — before the error replay, before `break`, and before the cleanup block:

```
$ bash -c 'set -e; DRAWN=3
clearProgress() { if [ "$DRAWN" -gt 0 ]; then printf "x" >/dev/full; DRAWN=0; fi; }
false || { clearProgress; echo "ERROR REPLAY REACHED"; }
echo "STILL ALIVE"'
printf: write error: No space left on device
rc=1                      # neither line printed
```

Round three had identical exposure at the redraw site (the `printf` was inline in the same spot), so this isn't new *there* — but the failure path previously was a bare `|| break` with no `printf` at all, so that path newly acquires it.

Reachability is thin: the branch requires `[ -t 1 ]`, so this needs a write to a terminal to fail, i.e. `EIO` after the pty is gone — and that normally arrives with `SIGHUP`, which kills the script anyway. I could not construct a realistic case. Mentioning it only because it's the same shape as the `set -e`-skips-cleanup bug round two fixed; `printf ... || true` would close it for good if you want the belt and braces. Fine to leave.

### Reflow-on-resize

Agreed with your call, no change requested — still bounded, still self-correcting, and avoiding it means the alternate screen you're removing.

---

cprussin:
Took the belt-and-braces in `a326f2b`: `printf '\033[%dA\033[J' "$DRAWN" || true`. Agreed it's barely reachable, but the failure mode it guards against is exactly the one round two was about — `set -e` skipping cleanup and leaving the pool imported with the LUKS device unlocked — so the cost/benefit is one word.

Re-verified after the change at 60x24: normal completion clean with cleanup reached, failure-at-poll-3 clean with cleanup reached, and zero escape sequences when piped.

Good catch on the successful-poll-stderr case that `2>&1` also fixes — I'd only reasoned about the failure path. Both behaviour changes you noted are the ones I'd want: a warning repeated every second for hours is worse than one at the end, and it's part of the status text at that point, so stdout is the right stream for it.

---

cprussin:
_Posted as a plain comment: the GraphQL review endpoint is still rate-limited._

**Round 5 (final confirmation) — clean, nothing remains.**

Scope: `git diff 6db613f..a326f2b`. Confirmed it is exactly one line — ` || true` appended to the `printf` in `clearProgress`. Word-diff and a byte-level file comparison show no other edit anywhere in `config/machines/crux/backup.nix`, and no other file is touched (`1 file, 1 insertion, 1 deletion`).

Verified empirically (bash under a pty):

- `clearProgress` returns 0 in all three states — `if` not taken (`rc=0`), taken with `printf` succeeding (`rc=0`, `DRAWN` reset), taken with `printf` failing (`rc=0`, `DRAWN` reset to 0). The `DRAWN=0` assignment still runs after a failed `printf`.
- Failure path, full loop simulation with a failing `printf` inside the taken branch:
  - **with** `|| true` → survives, error replay reaches stderr, `break` fires, cleanup block reached, exit 0;
  - **without** `|| true` → aborts under `set -e` at exit 1, no replay and no cleanup.
  
  This is the exact regression the hardening closes: the `{ clearProgress; echo "$NEXT" >&2; break; }` group is on the right of `||`, so `set -e` is *not* suppressed inside it and a non-zero `clearProgress` there would have killed the script before the pool was exported.
- Non-tty path unchanged: `DRAWN` stays 0, the drawing block never runs, `clearProgress` no-ops, error replay and cleanup still reached, exit 0.
- Nix `''` escaping unaffected: the added text contains no `'`, `$`, or `\`. Token census is identical across both commits (31 `${`, 6 `''`, 1 `''${`, 20 `\`), and the extracted script passes `bash -n` at both revisions with the one-line shell diff as the sole difference.

Fresh-eyes pass over the whole scrub section: the `exec 3>&1` / `tput cols <&3` save-and-restore, the `cut -c "1-$WIDTH"` truncation that keeps the drawn-line count equal to the physical rows, the `STATUS`-only-on-success replay, and the `exec 3>&-` teardown all still read correctly. No blocking or non-blocking defects found. Ship it.

---

PR #13, opened 2026-08-24
[ci] Add GitHub Actions CI

cprussin:
Adds `.github/workflows/ci.yml`, running on PRs, pushes to `main`, and on demand.

## Jobs

| Job | What it does |
| --- | --- |
| `cli test` | `nix develop -c cli test {lint,dead-code,format}` — statix, deadnix, alejandra, each as its own step so a failure is obvious |
| `eval lyra` / `eval crux` | Instantiates `colmenaHive.nodes.<m>.config.system.build.toplevel` |
| `eval installer` / `eval gpg-offline` | Instantiates the iso packages |
| `build pkgs.*` | Realizes this repo's own derivations: `dircolors-solarized`, `emojione-png`, `mako`, `notify-send`, `pass`, `powerpanel`, `sudo` |
| `build iot-devices` | `nix build .#iot-devices` |
| `build <iso> iso` | Full `nix build` of each iso — **`workflow_dispatch` only** |

A PR run is about 3 minutes end to end.

## Notes

- **CI always evaluates the locked inputs.** Updating them is its own change, so there's no `nix flake update` anywhere in here.
- **Machines are instantiated, not realized.** `lyra` can't be built anywhere but your machine: `displaylink` is a `requireFile`, so the source has to be added to the store by hand. A full `crux` closure also doesn't fit on a runner. Instantiating the toplevel still evaluates every module, option and overlay, which is what actually breaks when this repo changes.
- **Evaluation is pure**, matching `colmena apply` (which sets `--pure-eval` for flake hives). `allowUnfree` and `permittedInsecurePackages` come from the `nixpkgs.config` both machines get via `profiles/base` — colmena passes `meta.nixpkgs`'s config in as a module option rather than setting `nixpkgs.pkgs`, so the repo's own config applies.
- **`build-pkgs` reaches the overlays through `colmenaHive.nodes.lyra.pkgs.<attr>`**, so no flake restructuring was needed. Instantiation computes fixed-output hashes without fetching and runs no build phase, so without this a stale `src` hash or a broken `installPhase` would evaluate clean and first surface on `colmena apply`.
- **`emacs`, `google-fonts` and `launcher` are excluded** from that matrix — each realizes a multi-gigabyte closure (`launcher` interpolates every browser, `freecad`, `prusa-slicer`…) to check a handful of scripts; `launcher` alone took 549s and tripled the run. Their interpolations are still checked by eval.
- **One fix outside the workflow**: `cli.nix` now runs `deadnix --fail`. Without the flag deadnix reports findings and exits 0, so `cli test dead-code` could never fail — locally or in CI.
- **No `nix flake check`.** It adds nothing over these jobs (there are no `checks` outputs). Separately: `packages.<system>` maps `isos.<name>."${system}"`, which only exists for `x86_64-linux`, so `nix flake show` and `nix flake check --all-systems` do break today. Worth fixing in the flake, but out of scope here.
- **No binary-cache/store caching.** Everything substitutes from `cache.nixos.org`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,148 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  # Only pull requests cancel: a superseded PR run is wasted work, but a run on
+  # `main` is the record of whether that commit is good.
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+permissions:
+  contents: read
+
+jobs:
+  cli-test:
+    name: cli test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - uses: cachix/install-nix-action@v31
+      - name: Build the dev shell
+        run: nix develop --print-build-logs --command true
+      - name: cli test lint
+        if: ${{ !cancelled() }}
+        run: nix develop --command cli test lint
+      - name: cli test dead-code
+        if: ${{ !cancelled() }}
+        run: nix develop --command cli test dead-code
+      - name: cli test format
+        if: ${{ !cancelled() }}
+        run: nix develop --command cli test format
+
+  build:
+    name: build ${{ matrix.package }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        package:
+          - iot-devices
+    steps:
+      - uses: actions/checkout@v5
+      - uses: cachix/install-nix-action@v31
+      - name: Build ${{ matrix.package }}
+        run: nix build --print-build-logs ".#${{ matrix.package }}"
+
+  # Machine configurations can't be realized here: `lyra` pulls in `displaylink`,
+  # whose source is a `requireFile` that has to be fetched by hand, and a full
+  # `crux` closure is far bigger than a runner's disk.  Instantiating the
+  # toplevel derivation still evaluates every module, option and overlay, which
+  # is what actually breaks when this repo changes.
+  eval-machines:
+    name: eval ${{ matrix.machine }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        machine:
+          - lyra
+          - crux
+    steps:
+      - uses: actions/checkout@v5
+      - uses: cachix/install-nix-action@v31
+      - name: Instantiate ${{ matrix.machine }}
+        # Pure, like `colmena apply`, which evaluates flake hives with
+        # `--pure-eval`.  `allowUnfree` and `permittedInsecurePackages` come
+        # from the `nixpkgs.config` both machines get via `profiles/base`.
+        run: |
+          nix eval --raw \
+            ".#colmenaHive.nodes.${{ matrix.machine }}.config.system.build.toplevel.drvPath"
+          echo
+
+  eval-isos:
+    name: eval ${{ matrix.iso }} iso
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        iso:
+          - installer
+          - gpg-offline
+    steps:
+      - uses: actions/checkout@v5
+      - uses: cachix/install-nix-action@v31
+      - name: Instantiate the ${{ matrix.iso }} iso
+        run: |
+          nix eval --raw ".#packages.x86_64-linux.${{ matrix.iso }}.drvPath"
+          echo
+
+  # The derivations in `pkgs/` are only reached through the machines' overlays,
+  # and instantiating a machine runs no build phase -- a broken install phase or
+  # a stale `src` hash evaluates clean and first breaks on deploy.  Build them
+  # through `lyra`'s package set, which has every overlay in this repo applied.
+  # `emacs`, `google-fonts` and `launcher` are left out: each realizes a
+  # multi-gigabyte closure (`launcher` interpolates every browser, `freecad`,
+  # `prusa-slicer`...) to check a handful of scripts, and took 3x the rest of
+  # the workflow combined.  Their interpolations are still checked by eval.
+  build-pkgs:
+    name: build pkgs.${{ matrix.package }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        package:
+          - dircolors-solarized
+          - emojione-png
+          - mako
+          - notify-send
+          - pass
+          - powerpanel
+          - sudo
+    steps:
+      - uses: actions/checkout@v5
+      - uses: cachix/install-nix-action@v31
+      - name: Build ${{ matrix.package }}
+        run: |
+          nix build --print-build-logs \
+            ".#colmenaHive.nodes.lyra.pkgs.${{ matrix.package }}"
+
+  # Actually realizing an iso takes tens of minutes and most of a runner's disk,
+  # so it runs on demand only -- from the Actions tab or `gh workflow run
+  # ci.yml` -- rather than on every change.
+  build-isos:
+    name: build ${{ matrix.iso }} iso
+    if: github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        iso:
+          - installer
+          - gpg-offline
+    steps:
+      - uses: actions/checkout@v5
+      - name: Free up disk space
+        run: |
+          sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc \
+            /usr/local/share/boost /usr/local/share/powershell \
+            "${AGENT_TOOLSDIRECTORY:-/opt/hostedtoolcache}"
+          df -h /
+      - uses: cachix/install-nix-action@v31
+      - name: Build the ${{ matrix.iso }} iso
+        run: nix build --print-build-logs ".#${{ matrix.iso }}"
--- a/cli.nix
+++ b/cli.nix
@@ -14,7 +14,7 @@ lib.mkCli "cli" {
 
   test = {
     lint = "${statix}/bin/statix check .";
-    dead-code = "${deadnix}/bin/deadnix .";
+    dead-code = "${deadnix}/bin/deadnix --fail .";
     format = "${alejandra}/bin/alejandra --check .";
   };
 


.github/workflows/ci.yml
```diff
@@ -0,0 +1,137 @@
...
+        # when `nixpkgs.pkgs` is unset).  Deploys work locally because nixpkgs
+        # falls back to `~/.config/nixpkgs/config.nix`, which this repo installs
+        # from the same file; point `NIXPKGS_CONFIG` at it to get the same
+        # `allowUnfree` and `permittedInsecurePackages` here.
+        env:
+          NIXPKGS_CONFIG: ${{ github.workspace }}/config/modules/system/nix/nixpkgs-config.nix
+        run: |
+          nix eval --impure --raw \
+            ".#colmenaHive.nodes.${{ matrix.machine }}.config.system.build.toplevel.drvPath"
+          echo
```
cprussin: `--impure` + `NIXPKGS_CONFIG` shouldn't be needed here, and the rationale doesn't hold: colmena never sets `nixpkgs.pkgs`.

At the pinned rev (`dc22786`), `src/nix/hive/eval.nix` evaluates each node with `evalConfig` = `nixos/lib/eval-config.nix` and passes only `system`/`modules`/`specialArgs` — no `pkgs`. `meta.nixpkgs` is fed in purely as module options:

```nix
nixpkgs.overlays = lib.mkBefore npkgs.overlays;
nixpkgs.config = if hasTypedConfig then lib.mkBefore npkgs.config else lib.mkOptionDefault npkgs.config;
```

So `nixpkgs.pkgs` is undefined and the nixpkgs module re-imports nixpkgs from `cfg.config`/`cfg.overlays`. The `nixpkgs.config = import ./nixpkgs-config.nix` in `config/modules/system/nix/default.nix` — which both machines get via `config/profiles/base` — is what supplies `allowUnfree`/`permittedInsecurePackages`, in CI as well as locally.

Two supporting details:

- `nixpkgs.config` is not silently ignored when `nixpkgs.pkgs` *is* set; `nixos/modules/misc/nixpkgs.nix` (at the locked nixpkgs rev) asserts `opt.pkgs.isDefined -> cfg.config == { }`, so that combination is a hard eval error, not a fallback.
- The `~/.config/nixpkgs/config.nix` fallback can't be what makes local deploys work: for a flake hive colmena passes `--pure-eval` (`src/nix/hive/mod.rs:232`, `flags.set_pure_eval(self.path.is_flake())`), so `getEnv`/`$HOME` are empty there too.

Net effect as written: CI evaluates *less* strictly than `colmena apply` does, so if the module ever stopped providing `allowUnfree` CI would stay green while the deploy broke. Suggest:

```yaml
      - name: Instantiate ${{ matrix.machine }}
        run: |
          nix eval --raw ".#colmenaHive.nodes.${{ matrix.machine }}.config.system.build.toplevel.drvPath"
          echo
```

I couldn't run nix in this environment to confirm, but nothing under `config/`, `modules/` or `lib/` uses `getEnv`, `currentSystem`, `<nixpkgs>` or absolute-path imports, so pure eval should be fine. If it isn't, that's a real bug worth surfacing rather than working around.

(Unrelated, the attr path itself is correct — `nodes.<name>.config…` exists on `makeHive`'s output; `colmenaHive.toplevel.<name>.drvPath` is an equivalent shorthand.)

---

.github/workflows/ci.yml
```diff
@@ -0,0 +1,137 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+  # Catches breakage that comes from upstream rather than from a change here.
+  schedule:
+    - cron: "0 12 * * 1"
```
cprussin: The schedule won't catch upstream breakage as written: `flake.lock` pins every input, so the Monday run evaluates byte-identical inputs to the PR run. The only things it can newly catch are non-hermetic failures — e.g. the `gpg-key` input (a live `https://connor.prussin.net/pubkey.asc` fetch, re-fetched on every fresh runner) going away, or a substituter miss.

To get what the comment promises, the scheduled run needs to update inputs first, e.g. a `nix flake update` (or `--override-input nixpkgs github:NixOS/nixpkgs/nixos-26.05`) step gated on `github.event_name == 'schedule'`.

---

.github/workflows/ci.yml
```diff
@@ -0,0 +1,137 @@
...
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
cprussin: `cancel-in-progress` also applies to `main`, and scheduled runs report `github.ref` as `refs/heads/main` too — so a weekly run and a push to `main` share the group `CI-refs/heads/main`. Any merge landing during the Monday run cancels it, and the replacement (push) run skips `build-isos` because its `github.event_name` is `push`. The only job that ever realizes an iso can therefore be dropped silently.

Easiest fix is to scope cancellation to PRs:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

(or add `${{ github.event_name }}` to the group).

---

.github/workflows/ci.yml
```diff
@@ -0,0 +1,137 @@
...
+jobs:
+  cli-test:
+    name: cli test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - uses: cachix/install-nix-action@v31
+        with:
+          extra_nix_config: |
+            experimental-features = nix-command flakes
```
cprussin: Minor: this is a no-op. `install-nix-action`'s `install-nix.sh` already writes it when `extra_nix_config` doesn't mention it:

```sh
if [[ ! $INPUT_EXTRA_NIX_CONFIG =~ "experimental-features" ]]; then
  add_config "experimental-features = nix-command flakes"
```

The whole `with:` block can be dropped from all four jobs (`uses: cachix/install-nix-action@v31` on its own), removing the duplication.

---

cprussin (commented):
Reviewed the workflow against the flake, the pinned colmena (`dc22786`) and nixpkgs (`a9e6d84`) sources, and `mkCli`. Structure and job split look right; four comments inline, one of them substantive.

**Main point:** the `--impure` + `NIXPKGS_CONFIG` in `eval-machines` rests on an incorrect premise — colmena doesn't set `nixpkgs.pkgs`, so the repo's own `nixpkgs.config` is what applies, both in CI and in `colmena apply` (which uses `--pure-eval` for flake hives). Details inline; this also affects the "pass `config` to `meta.nixpkgs`" follow-up idea in the PR description, which isn't needed either.

**Verified as correct** (no action needed):

- Attr paths: `colmenaHive.nodes.<name>.config.system.build.toplevel` exists on `makeHive`'s output; `packages.x86_64-linux.{installer,gpg-offline}` match `mkMinimalIsos`, and `hydraJob` (which `release.nix`'s `makeIso` wraps the iso in) preserves `drvPath`, so the `.drvPath` eval works. `iot-devices` matches the overlay.
- `nix develop --command cli test <x>`: `mkCli` emits a plain `case` dispatch and the leaf command's status is the script's status, so failures propagate. (`cli test` on its own would run all three in parallel, but separate steps give better per-check reporting — fine as is.)
- YAML parses; `if: ${{ !cancelled() }}`, the matrices and the `build-isos` event gate are all valid; `actions/checkout@v5` and `cachix/install-nix-action@v31` both exist as tags.
- No rate-limit risk: the action configures `access-tokens = github.com=$GITHUB_TOKEN` from `github.token` automatically.
- Eval jobs should not hit unfree/insecure errors: both machines import `config/modules/system/nix` via `config/profiles/base`, and the isos pull in `hardware.enableAllHardware` only (not `enableAllFirmware`), so they evaluate under the default config the same way `nix build .#installer` does locally.
- `displaylink` is a `requireFile`, which only fails at build time, so instantiating `lyra` is fine.

**One correction to the PR description:** `nix flake check` only checks the current system unless `--all-systems` is passed, so the `isos.<name>."${system}"` gap wouldn't break a plain `nix flake check` on x86_64-linux. It still adds little over the jobs here (there are no `checks` outputs), so skipping it is reasonable — just not for that reason. The underlying `packages.<non-x86_64>` breakage is real for `nix flake show`/`--all-systems` and worth fixing in the flake at some point.

---

.github/workflows/ci.yml
cprussin: Confirmed and fixed — dropped `--impure` and `NIXPKGS_CONFIG`.

I re-checked both sources independently rather than taking this on trust:

- `src/nix/hive/eval.nix` at `dc22786`: `evalConfig { inherit (npkgs.stdenv.hostPlatform) system; modules = [ nixpkgsModule … ]; }` — no `pkgs`, and `nixpkgsModule` feeds `meta.nixpkgs`'s config in as `lib.mkBefore npkgs.config`. So `nixpkgs.pkgs` is undefined and the module's `nixpkgs.config` is what's in effect.
- `nixos/modules/misc/nixpkgs.nix` at `a9e6d84`: `finalPkgs = if opt.pkgs.isDefined then cfg.pkgs.appendOverlays cfg.overlays else defaultPkgs`, plus `assertion = opt.pkgs.isDefined -> cfg.config == { }`. My "config is silently ignored" reading came from the option's *description*, which says other `nixpkgs.*` options "will be ignored" — the assertion is the operative behavior.

So the original rationale was backwards, and worse, it made CI evaluate more permissively than a deploy. Pure eval now, matching `colmena apply`. PR description updated too.

---

.github/workflows/ci.yml
cprussin: Right — the lock made that comment a lie. Added a `nix flake update` step gated on `github.event_name == 'schedule'` to `cli-test`, `build`, `eval-machines` and `eval-isos`.

Deliberately *not* added to `build-isos`: realizing an iso against a moved nixpkgs means a from-source rebuild of most of the closure, which won't finish on a runner. That job stays on the locked inputs, which is noted in its comment.

---

.github/workflows/ci.yml
cprussin: Good catch — a merge silently eating the only iso-building run is exactly the failure mode that would go unnoticed. Applied your suggestion: `cancel-in-progress: ${{ github.event_name == 'pull_request' }}`, with the reason in a comment.

---

.github/workflows/ci.yml
cprussin: Verified against `install-nix.sh` at the `v31` tag (`if [[ ! $INPUT_EXTRA_NIX_CONFIG =~ "experimental-features" ]]; then add_config …`) — it is a no-op. Dropped the `with:` block from all four jobs.

---

cprussin:
## Review round 2

Re-read `.github/workflows/ci.yml` fresh against the flake, the pinned nixpkgs (`d407951`), colmena (`dc22786`), `mkCli`, the `install-nix-action` source, and the logs from run [32746276233](https://github.com/cprussin/dotfiles/actions/runs/32746276233). One real defect, one gap in the scheduled path, and two notes.

> Posting as a single comment rather than inline review comments — the GraphQL endpoint the review API needs has been rate-limited for this account for the last couple of hours. Line references are given inline below.

---

### 1. `cli test dead-code` can never fail — `.github/workflows/ci.yml:38-40`

This step is an inert gate.

`cli.nix` defines `test.dead-code = "${deadnix}/bin/deadnix ."`. The pinned nixpkgs (`d407951447dcd00442e97087bf374aad70c04cea`) ships deadnix **1.3.1**, and in that version's `src/main.rs` the *only* non-zero exit is gated behind `--fail`:

```rust
let fail_on_reports = matches.get_flag("FAIL_ON_REPORTS");
// ...
if fail_on_reports && report_count > 0 {
    std::process::exit(1);
}
```

`-f, --fail  Exit with 1 if unused code has been found`. Without it, deadnix prints its findings on stdout and exits **0**. `mkCli` propagates the exit code correctly (`exit ${PIPESTATUS[0]}`), so nothing else is swallowing it — deadnix simply never signals failure.

Nothing is being masked *today*: in the run on this head the step took 0.2s and emitted no output at all, so the tree is currently clean. But the workflow advertises a check it does not perform, and the first time dead code lands it will go green.

For contrast, the other two steps are genuine gates, which is why this one stands out:

- statix (`0.5.8-unstable-2026-06-28`, `bin/src/lint.rs`): `std::process::exit(1)` when `!results.is_empty()`.
- alejandra 4.0.0: documented `2, if --check was used and any file requires formatting`, and `std::process::exit(if in_place { 0 } else { 2 })`.

Fix belongs in `cli.nix`, not in the workflow:

```nix
test = {
  lint = "${statix}/bin/statix check .";
  dead-code = "${deadnix}/bin/deadnix --fail .";   # <-- add --fail
  format = "${alejandra}/bin/alejandra --check .";
};
```

Leave `fix.dead-code` on `-e` as-is. Worth folding into this PR since this PR is what turns that command into a CI gate.

---

### 2. The scheduled `nix flake update` leaves no record of what it resolved — `ci.yml:30-32`, `56-58`, `78-80`, `103-105`

The weekly job's whole purpose is to catch upstream breakage, but when it goes red you get an evaluation error and nothing to attribute it to — no lock diff, no revisions, nothing to bisect against. Worth adding to each update step (needs nothing beyond the existing `contents: read`):

```yaml
      - name: Update flake inputs
        if: github.event_name == 'schedule'
        run: |
          nix flake update
          git --no-pager diff -- flake.lock
```

Secondary, and lower probability: `nix flake update` runs **independently in four jobs**, so each resolves `nixpkgs`, `nixpkgs-unstable`, `colmena` etc. at its own moment. In practice the jobs start within seconds of each other so they'll almost always agree, but nothing guarantees it — a scheduled run can in principle be green in `eval lyra` and red in `cli test` against two different revisions of the same branch. Resolving once in a small setup job and passing `flake.lock` to the rest via `actions/upload-artifact` / `download-artifact` makes the run atomic *and* gives you the record above for free.

Also worth noting: the updated lock is discarded entirely, so a green weekly run doesn't leave you anything to land — the actual bump is still a manual follow-up. Uploading the lock as an artifact would at least make it copy-pasteable.

---

### 3. Coverage gap: this repo's own derivations are instantiated but never built

Answering the "is anything worth testing that CI doesn't touch" question directly.

Everything under `pkgs/` — `emacs-rc`, `launcher`, `mako`, `powerpanel`, `pass-with-otp`, `sudo-with-insults`, `emojione-png`, `google-fonts-without-noto-emoji`, `zoom-frm`, `notify-send`, `dircolors-solarized` — is reached only through `nixpkgs.overlays` in `config/modules/system/nix`, `config/modules/ui/{emacs,launcher,fonts}` etc. `eval-machines` forces `.drvPath` on the toplevel, which recursively instantiates all of them, but instantiation computes fixed-output hashes **without fetching** and never runs a build phase. So a stale `sha256` on a `src`, or a broken `buildPhase`/`installPhase`/`postPatch`, passes CI clean and first surfaces on the next `colmena apply`. The weekly iso build only covers the couple the isos pull in (`dircolors-solarized`).

Meanwhile `build iot-devices` — the one job that actually realizes something on PRs — is `linkFarmFromDrvs` over `writeText`. It completed in 12s and catches essentially nothing that eval doesn't already catch.

Not a blocker, and it isn't free (several of these are `override`s that rebuild the underlying package from source — `sudo`, `mako`), so it's a real tradeoff rather than an obvious win. But if the `build` matrix is going to exist with a single entry, exposing the `pkgs/` derivations as flake `packages` and adding the cheap ones to that matrix is the highest-value thing you could add to it.

---

### 4. Two smaller notes

**`build-isos` has never actually executed** (`ci.yml:116-136`). It's skipped on `pull_request`, and `workflow_dispatch` can't target a workflow that isn't on the default branch yet — so the first real execution will be post-merge. The cleanup list itself is fine (`AGENT_TOOLSDIRECTORY` is set to `/opt/hostedtoolcache` on hosted runners and nothing later in the job needs it), but if a full `nix build .#installer` doesn't fit after the cleanup, you'll learn that from a red `main` rather than from this PR. Probably worth a manual dispatch right after merge instead of waiting for Monday.

**On the "no binary-cache/store caching" note in the PR description** — one specific thing to know before you go looking. In run 32746276233 the `cli test` job spent 1m07s of its 1m48s compiling colmena with rustc, for three linters that don't touch colmena. The obvious fix does *not* work: adding `https://colmena.cachix.org` as a substituter will miss, because this flake pulls colmena in via `colmena.overlays.default`, which is

```nix
overlays.default = final: prev: {
  colmena = final.callPackage ./package.nix { };
};
```

i.e. it rebuilds against *this* repo's nixpkgs (`nixos-26.05`), so the output hashes don't match the cache's artifacts (built against colmena's own `nixpkgs-unstable`). Hitting that cache would mean using `colmena.packages.${system}.colmena` in the dev shell instead of the overlay — a flake change with its own tradeoffs. Caching `/nix/store` across runs is the other route. Entirely optional; 2 minutes per PR may just be fine.

---

### Verified clean (no action needed)

- `if: ${{ !cancelled() }}` behaves as intended — a prior step's *failure* does not set `cancelled()`, so all three lint steps run and report independently.
- `mkCli` exit-code propagation: `${cmd} "$@" 2>&1 | sed ...` followed by `exit ${PIPESTATUS[0]}`, under `writeShellScriptBin` (bash), so `PIPESTATUS` is available. Failures do reach the step.
- `install-nix-action@v31` needs no `github_access_token`: `install-nix.sh` falls back to `access-tokens = github.com=$GITHUB_TOKEN` from the `github.token` it receives via `env`, so no unauthenticated rate-limit exposure on the ~12 GitHub flake inputs.
- Pure eval genuinely works on a dirty tree here — nothing in this repo reads `self.rev` or sets `system.configurationRevision`, so `nix flake update` + `nix eval` on a modified lock can't trip the dirty-tree restriction.
- `concurrency` expression, the `build-isos` event gate, both matrices, `actions/checkout@v5`, and the `.#packages.x86_64-linux.<iso>.drvPath` / `.#colmenaHive.nodes.<m>...drvPath` attribute paths all check out; the run on this head confirms it end to end.
- `nix build` drops `./result`, which `.gitignore` already covers, so no dirty-tree side effects between steps.

---

cprussin:
Round 2 addressed. Verified each claim against source before acting.

**1. `cli test dead-code` was an inert gate — fixed in `cli.nix`.**

Confirmed against deadnix's `src/main.rs`: the only non-zero exit is `if fail_on_reports && report_count > 0`, gated on `-f/--fail`. Without it, deadnix reports and exits 0. Changed `test.dead-code` to `deadnix --fail .` (left `fix.dead-code` on `-e`). This is the right call regardless of CI — the check was already a no-op locally.

(Minor: the review cites nixpkgs `d407951`, but this flake's `nixpkgs` input locks to `a9e6d84`. Doesn't affect the finding — the `--fail` gating is upstream behavior, not version-specific.)

**2. Lock diff on scheduled runs — added.** Each `Update flake inputs` step now runs `git --no-pager diff -- flake.lock` after the update, so a red weekly run says which revisions it moved to.

Skipped the resolve-once-and-pass-via-artifact suggestion. It's correct that four independent `nix flake update` calls aren't atomic, but the window is seconds and the artifact plumbing costs more complexity than the race is worth here. Easy to add if a weekly run ever comes back inconsistent.

**3. Coverage gap — fixed, new `build-pkgs` job.**

This was the most useful finding: instantiation computes fixed-output hashes without fetching and runs no build phase, so a broken `installPhase` in `pkgs/` evaluated clean. The job builds them through `.#colmenaHive.nodes.lyra.pkgs.<attr>` — `lyra` (laptop → pc → base) has every overlay in this repo applied, so no flake restructuring is needed to reach them:

`dircolors-solarized`, `emojione-png`, `launcher`, `mako`, `notify-send`, `pass`, `powerpanel`, `sudo`

Left out deliberately: `emacs` (a `symlinkJoin` over a full emacs + elisp tree) and `google-fonts` (a multi-gigabyte fetch to delete two files) — the build cost dwarfs what building them checks. `zoom-frm` lives inside `emacsPackagesFor` and comes with `emacs`.

**4. Notes — acknowledged, no change.**

- `build-isos` genuinely can't run until this is on `main`. I'll dispatch it manually after merge rather than finding out on a Monday.
- The `colmena.cachix.org` dead end is useful to have written down; the overlay does rebuild against this repo's nixpkgs. Not pursuing it — 2 minutes per PR is fine.

CI is running on the new head; I'll confirm the `pkgs` builds are as cheap as they look and trim the matrix if any turns out not to be.

---

cprussin:
Follow-up on the `build-pkgs` matrix, from what the run on `97f0320` actually measured.

Everything came back green, and all but one build was cheap:

| | |
| --- | --- |
| `dircolors-solarized` | 35s |
| `emojione-png` | 42s |
| `mako` | 39s |
| `notify-send` | 40s |
| `pass` | 37s |
| `powerpanel` | 35s |
| `sudo` | 87s |
| **`launcher`** | **549s** |

`launcher` is an outlier by an order of magnitude, and for a bad reason: its scripts interpolate `chromium`, `google-chrome`, `brave`, `tor-browser`, `firefox`, `emacs`, `freecad`, `prusa-slicer`, `mpv`, `ffmpeg` and friends into shell strings, so realizing it drags in that entire closure. Several GB of substitution to validate six `writeShellScriptBin` scripts and one `ln -s` in `postBuild`, and it tripled the workflow's wall-clock on its own.

Dropped it from the matrix, alongside `emacs` and `google-fonts`, for the same reason. Those interpolations are still checked by `eval-machines` — that's what would catch a renamed or removed attribute. What's lost is only the `symlinkJoin`'s `postBuild`, which is one line.

Easy to add back if you'd rather have the coverage; it's one line in the matrix.

Also worth noting the `cli test` job stayed green with `deadnix --fail`, so the tree really is clean — the check wasn't just passing silently.

---

PR #14, opened 2026-08-25
[flake] Only expose the isos on systems they're built for

cprussin:
Fixes `nix flake show --all-systems` and `nix flake check --all-systems`, which throw today.

## The bug

`mkMinimalIso` passes `supportedSystems = ["x86_64-linux"]` to `release.nix`, so each entry in `isos` is an attrset with exactly one key:

```nix
isos.installer = { x86_64-linux = <drv>; };
```

`packages` then selected off it unconditionally:

```nix
packages = (builtins.mapAttrs (_: value: value."${system}") isos) // { ... };
```

`eachDefaultSystem` also runs for `aarch64-linux`, `x86_64-darwin` and `aarch64-darwin`, where that key doesn't exist — so `packages.aarch64-linux.installer` throws when forced.

Scope, measured rather than assumed:

| command | before |
| --- | --- |
| `nix flake show` | passes — prints `omitted (use '--all-systems' to show)`, never forces the other systems |
| `nix flake check` | passes — warns about omitted incompatible systems |
| `nix flake show --all-systems` | `error: attribute 'aarch64-darwin' missing` |
| `nix flake check --all-systems` | `error: attribute 'aarch64-linux' missing` |

`mapAttrs` is lazy in its values, which is why the plain forms were fine. `nix build .#installer` was never affected either — it only resolves the current system.

## The fix

Filter to the systems each iso actually has a build for, derived from the attrs themselves rather than by re-hardcoding `x86_64-linux`:

```nix
isoPackages =
  builtins.mapAttrs (_: value: value."${system}")
  (nixpkgs.lib.filterAttrs (_: builtins.hasAttr system) isos);
```

`packages.x86_64-linux` is unchanged — `installer`, `gpg-offline`, `iot-devices`. Other systems now expose just `iot-devices` instead of two attributes that throw.

## Not covered by this change

`devShells.default` is still evaluated per system and pulls `colmena` and `vulnix` (via `cli`). If those are Linux-only in `meta.platforms`, `nix flake check --all-systems` may still fail further along on darwin. Unverified — flagging it so `--all-systems` isn't assumed green on the strength of this PR alone.

Spotted while adding CI in #13; it's why that workflow builds specific attributes instead of running `nix flake check`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -148,9 +148,16 @@
             ];
             config = {};
           };
+
+          # The isos are only built for the systems in `supportedSystems`
+          # above, so drop the ones this system has no build for rather than
+          # leaving `packages` with attributes that throw on access.
+          isoPackages =
+            builtins.mapAttrs (_: value: value."${system}")
+            (nixpkgs.lib.filterAttrs (_: builtins.hasAttr system) isos);
         in {
           packages =
-            (builtins.mapAttrs (_: value: value."${system}") isos)
+            isoPackages
             // {
               inherit (pkgs) iot-devices;
             };


flake.nix
```diff
@@ -148,9 +148,16 @@
             ];
             config = {};
           };
+
+          # The isos are only built for the systems in `supportedSystems`
+          # above, so drop the ones this system has no build for rather than
+          # leaving `packages` with attributes that throw on access.
+          isoPackages =
+            builtins.mapAttrs (_: value: value."${system}")
+            (nixpkgs.lib.filterAttrs (_: builtins.hasAttr system) isos);
```
cprussin, line 157: Verified correct.

- **Currying.** `lib/attrsets.nix:667` at the pinned rev is `filterAttrs = pred: set: removeAttrs set (filter (name: !pred name set.${name}) (attrNames set));` — `pred` is applied as `pred name value`, so `(_: builtins.hasAttr system)` drops the name and applies `hasAttr system` to the value. Right shape.
- **`nixpkgs.lib` resolves here.** `nixpkgs` comes from the `outputs` argument list and isn't shadowed inside the per-system lambda (`pkgs` is a separate binding). Not just assumed: CI's `eval installer iso` on this head forces `packages.x86_64-linux`, which now goes through `filterAttrs`, and it passes — in pure eval, so `release-lib.nix`'s `system ? builtins.currentSystem` default is never demanded either.
- **Laziness.** `filterAttrs` forces each `isos.<name>` to WHNF, which plain `mapAttrs` did not. Cheap in practice: `forAllSystems = genAttrs supportedSystems` builds the attrset spine without forcing the ISO derivations, and the `pkgs` bindings in both `release.nix` and `release-lib.nix` stay lazy. No meaningful eval cost added on the systems that now get nothing.

Optional, non-blocking: `["x86_64-linux"]` is now encoded twice — literally in `mkMinimalIso`, and implicitly here by probing for it. Hoisting it would make one source of truth:

```nix
isoSystems = ["x86_64-linux"];
# ... supportedSystems = isoSystems;
isoPackages =
  nixpkgs.lib.optionalAttrs (builtins.elem system isoSystems)
  (builtins.mapAttrs (_: value: value.${system}) isos);
```

That said, the probe is arguably the more robust of the two — it follows what `release.nix` actually produced rather than asserting what it should have. Take it or leave it.

---

cprussin (commented):
The fix is correct. One inaccuracy in the commit message worth amending; no code changes needed.

## Verified

Checked against the pinned nixpkgs (`a9e6d84f9c2f9012f5fe7d964a7851352300e61a`) and `flake-utils` (`11707dc2`), and by evaluating a standalone repro of both the old and new expressions with Nix 2.24.9.

- **Diagnosis is right.** `nixos/release.nix` defines `iso_minimal = forAllSystems (...)`, and `pkgs/top-level/release-lib.nix:159` defines `forAllSystems = genAttrs supportedSystems`. With `supportedSystems = ["x86_64-linux"]`, each `isos.<name>` is exactly `{ x86_64-linux = <drv>; }` — one key, nothing else. `flake-utils` at the pinned rev has `defaultSystems = ["aarch64-linux" "aarch64-darwin" "x86_64-darwin" "x86_64-linux"]`, so three of four had no key and the old select genuinely threw `attribute '<system>' missing`. Not harmless.
- **`packages.x86_64-linux` is unchanged** — `installer`, `gpg-offline`, `iot-devices`, same three attributes before and after. Other systems drop to `iot-devices` alone.
- **Nothing else has the same bug.** The only other places touching `isos` are `mkMinimalIsos` — whose `isos` parameter is a list of names, not the attrset, so there's no selection to get wrong — and `.github/workflows/ci.yml`, which uses `.#packages.x86_64-linux.<iso>` and `.#<iso>`. Both `x86_64-linux`, both unaffected.

Currying, `nixpkgs.lib` scoping, and the laziness change are covered in the inline comment.

## Worth fixing: the commit message overstates the symptom

Both the commit message and the PR body say this breaks `nix flake show`. It doesn't — only the `--all-systems` forms broke. Measured on the old expression:

| command | old behavior |
| --- | --- |
| `nix flake show` | **passes** — prints `omitted (use '--all-systems' to show)` for the other systems, never forcing them |
| `nix flake check` | **passes** — `warning: The check omitted these incompatible systems: aarch64-darwin, aarch64-linux, x86_64-darwin` |
| `nix flake show --all-systems` | `error: attribute 'aarch64-darwin' missing` |
| `nix flake check --all-systems` | `error: attribute 'aarch64-linux' missing` |

`mapAttrs` is lazy in its values, so the plain forms list the attribute *names* without ever forcing them. Both `--all-systems` forms pass after the change.

Suggest amending to `nix flake show --all-systems` in the commit message and PR body. The fix is right either way, but the commit message is the durable record of why it exists.

## One caveat on the stated goal

"Fixes `nix flake check --all-systems`" holds for `packages`, but `devShells.default` is still evaluated per system and pulls `pkgs.colmena`, `pkgs.vulnix` (via `cli`) and friends on `aarch64-darwin` / `x86_64-darwin`. I could not run this here to confirm, so I'm flagging it rather than asserting it — but if those have Linux-only `meta.platforms`, `checkMeta` throws at eval time and `--all-systems` would still fail, just further along. Doesn't affect the merit of this change; only worth knowing before treating `nix flake check --all-systems` as green.

---

flake.nix
cprussin: Thanks for actually measuring rather than reasoning about it — the laziness point is the one I got wrong.

**Amended.** Commit message and PR body now say `nix flake show --all-systems` / `nix flake check --all-systems`, with the four-way table showing the plain forms passed. I'd claimed plain `nix flake show` was broken in three places (here, #13's description, and in chat); `mapAttrs` being lazy in its values means it never forced the other systems. Corrected in all of them.

**Leaving the probe as-is.** You made the case against your own suggestion better than I would have: following what `release.nix` actually produced beats asserting what it should have. If `supportedSystems` ever grows, the probe adapts and the hoisted list would need remembering to update.

**Kept the `devShells` caveat**, marked unverified, in the PR body — worth having written down so `--all-systems` isn't assumed green on the strength of this change.

---

PR #16, opened 2026-08-25
[ci] Check the flake on every system

cprussin:
Adds a `flake-check` job running `nix flake show --all-systems` and `nix flake check --all-systems`.

## Why

This is the gap that let #14's bug sit unnoticed: `packages.<system>` carried iso attributes that threw when forced, on the three systems with no iso build. Nothing in CI would have caught it — and the `eval-isos` job that existed at the time couldn't have, since it pinned `x86_64-linux`, the one system that worked.

`--all-systems` is the load-bearing part. Without it both commands only look at the runner's own system.

Cheap job: `nix flake check` builds `checks` outputs (there are none here) and only evaluates everything else, so this is an eval-only pass.

## This may go red, and that would be the finding

The review on #14 flagged, unverified, that `devShells.default` is evaluated per system and pulls `colmena` and `vulnix` (via `cli`), plus `cli` reaches `esphome` through `iot.nix`. If any of those are Linux-only in `meta.platforms`, `checkMeta` throws on `x86_64-darwin` / `aarch64-darwin` and this job fails.

I haven't been able to run nix in this session, so this PR's own CI is the first real test. If it's red on darwin that's a genuine result, not a broken job: it means `flake-utils.lib.eachDefaultSystem` is advertising four systems this flake only supports one of. The fix would be narrowing the systems list rather than deleting the check — I'd raise it rather than merge around it.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JVNS9yNwURwF5tW2whza94)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,30 @@ jobs:
         if: ${{ !cancelled() }}
         run: nix develop --command cli test format
 
+  # `--all-systems` is the point: without it both commands only look at the
+  # runner's own system, which is the one that always worked.  It was the other
+  # three that carried outputs which threw when forced.
+  #
+  # `check` subsumes `show` for coverage -- it forces each `packages.<system>.*`
+  # to its `drvPath`, where `show` forces only far enough to print a name -- so
+  # the `show` step is here for its readable tree, not to catch anything `check`
+  # would miss.
+  flake-check:
+    name: flake check
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v5
+      - uses: cachix/install-nix-action@v31
+      - name: nix flake show
+        run: nix flake show --all-systems
+      - name: nix flake check
+        if: ${{ !cancelled() }}
+        # No `checks` outputs today, so this only evaluates.  Keep the build
+        # logs on rather than passing `--no-build`: if `checks` are ever added,
+        # they should actually run here, and say why when they fail.
+        run: nix flake check --print-build-logs --all-systems
+
   build:
     name: build ${{ matrix.package }}
     runs-on: ubuntu-latest


.github/workflows/ci.yml
```diff
@@ -35,6 +35,22 @@ jobs:
...
+  # three that carried outputs which threw when forced.
+  flake-check:
+    name: flake check
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v5
+      - uses: cachix/install-nix-action@v31
+      - name: nix flake show
+        run: nix flake show --all-systems
```
cprussin, line 54: `nix flake show --all-systems` adds no coverage over the `nix flake check --all-systems` below it, for this flake. From this run's job log:

- Both reach the same three outputs. `show` walked `devShells`, `overlays`, `packages`, and for `colmena`/`colmenaHive` printed `unknown` without recursing — a shallow force of a lazy attrset, so neither `meta.nixpkgs` nor any machine module gets evaluated.
- Per leaf, `check` forces strictly more. `show` forces a leaf far enough to render `installer: package 'installer.iso'` (is-it-a-derivation + `name`); `check` forces `drvPath`.
- `check` does one thing `show` doesn't: `checking overlay 'overlays.iot-devices'...`, which applies the overlay. `show` only tags it `Nixpkgs overlay`.

So `check` strictly subsumes `show` here, including for 09463c7's bug: `checking derivation packages.aarch64-linux.installer` would have forced the throwing `value."aarch64-linux"` selection on its own, without the `show` step existing.

The cost isn't nil. The two commands don't share an eval cache, so each instantiates both isos independently — `gpg-offline` 23s + `installer` 19s under `show`, then 23s + 22s again under `check`. The `show` step is 53s of the job's 129s.

Suggest dropping the step. If you want the tree printed as a human-readable artifact, that's a fine reason to keep it, but say so — as written the comment above the job justifies `--all-systems` and never says why there are two commands.

---

.github/workflows/ci.yml
```diff
@@ -35,6 +35,22 @@ jobs:
...
+      - name: nix flake check
+        if: ${{ !cancelled() }}
+        run: nix flake check --all-systems
```
cprussin: No `--no-build`. Correct today: the flake has no `checks` output, and the log confirms it — `colmena` → `overlays` → `packages` → `devShells` → `colmenaHive`, no `checks` phase. So this is eval-only, as the PR description says.

The scenario where that changes is not remote. `checks.<system>.*` is the natural home for the `lint` / `dead-code` / `format` targets `cli-test` currently runs by hand, and adding one silently converts this job into a build job — building every check on all four systems, i.e. cross-instantiating `checks.aarch64-darwin.*` on an x86_64-linux runner. Two things make that land worse than it needs to:

- Every other nix invocation in this file passes `--print-build-logs` (`nix develop` in `cli-test`, all three `nix build` steps). This one doesn't, so the first surprise build fails with no build log.
- `timeout-minutes: 30` is the only bound, and the job's own comment presents it as an eval pass, so the next reader won't expect builds here.

Cheapest fix that keeps the job's stated contract: `nix flake check --all-systems --no-build`. If you'd rather have checks genuinely built here once they exist, add `--print-build-logs` instead and adjust the comment.

---

cprussin (commented):
Reviewed the diff against `flake.nix`, the history around 09463c7, and the actual `flake check` job log for c0de6e2 (run 32812256626, job 97693718075).

The job does what the commit message claims. `--all-systems` is load-bearing, and `nix flake check --all-systems` forces `drvPath` on every `packages.<system>.*` — exactly what 09463c7's bug tripped over. Two findings inline: the `nix flake show` step earns nothing that `nix flake check` doesn't already cover, and the missing `--no-build` is a latent trap. Neither is a correctness problem with the job as it stands.

**Checked and fine:**

- `if: ${{ !cancelled() }}` on the second step does what's intended. It replaces the implicit `success()` so `nix flake check` still runs when `nix flake show` fails, matching the pattern in `cli-test`. (It's also the only thing making the two-command arrangement coherent — worth keeping if the `show` step stays.)
- Placement, `timeout-minutes: 30`, `actions/checkout@v5`, `cachix/install-nix-action@v31`, comment-above-the-job style: all consistent with the other five jobs.
- The job comment is accurate. Without `--all-systems` both commands restrict to the runner's system, and per 09463c7 it was the other three whose iso attrs threw when forced. It just doesn't explain why there are two commands.

**The two `unknown flake output` warnings** — not merely cosmetic, but not a hole either. `nix flake check` skips `colmena` and `colmenaHive` outright (log: `checking flake output 'colmena'...` → `warning: unknown flake output 'colmena'` → straight on to `overlays`), so a machine-config regression genuinely would not be caught by this job. Coverage is intact elsewhere: `eval-machines` instantiates `colmenaHive.nodes.{lyra,crux}.config.system.build.toplevel`, `build-pkgs` builds through `colmenaHive.nodes.lyra.pkgs`, and the `colmena` output is the same `colmena-machines` attrset `colmenaHive` is made from, so nothing is left unevaluated. Worth being explicit that "all checks passed!" from this job means `packages` + `devShells` + `overlays` only, and that removing `eval-machines` later would drop machine coverage with CI still green.

**One thing this run doesn't prove:** the fix landed in 09463c7 before this job existed, so neither command has ever been observed failing on the bug it's meant to catch. For `nix flake check` that's fine by construction — it forces `drvPath` on every `packages.<system>.*`, and the pre-09463c7 `builtins.mapAttrs (_: value: value."${system}") isos` made three of those attrs throw on force. For `nix flake show` the catch is asserted, not demonstrated; another reason not to carry the step.

---

.github/workflows/ci.yml
cprussin: Your analysis is right and I'm not disputing it — `check` forces `drvPath` where `show` forces a name, plus it applies the overlay, so `show` catches nothing `check` wouldn't. The 53s-of-129s measurement makes the cost concrete too.

Keeping the step anyway, for a reason outside the diff: the user asked for `nix flake check` **and** `nix flake show` by name. Dropping half of an explicit request on my own judgement isn't mine to make, so I've documented the step's actual role instead of quietly justifying it as coverage:

```yaml
  # `check` subsumes `show` for coverage -- it forces each `packages.<system>.*`
  # to its `drvPath`, where `show` forces only far enough to print a name -- so
  # the `show` step is here for its readable tree, not to catch anything `check`
  # would miss.
```

That was the part you were actually right to push on — the previous comment justified `--all-systems` and left the two-command arrangement unexplained, so the next reader would have assumed both were pulling weight. Now the file says which one isn't. I've put the redundancy and the 53s to the user; if they'd rather have the seconds than the tree, dropping it is one line.

---

.github/workflows/ci.yml
cprussin: Good catch on the trap, but I went the other way on the fix — `--print-build-logs` rather than `--no-build`.

`--no-build` removes the surprise by making the job never run checks. That means the day someone moves `lint`/`dead-code`/`format` into `checks.<system>.*` — which, as you say, is exactly where they belong — this job reports "all checks passed!" while running none of them. A green check that silently stopped checking is a worse failure mode than a slow job, because nothing surfaces it.

So: keep it able to build, and fix the thing that would actually hurt — the missing log. It's now consistent with every other nix invocation in the file, and the comment records the reasoning:

```yaml
        # No `checks` outputs today, so this only evaluates.  Keep the build
        # logs on rather than passing `--no-build`: if `checks` are ever added,
        # they should actually run here, and say why when they fail.
```

The cross-instantiation concern you raise is real and stays real under this choice — `checks.aarch64-darwin.*` on an x86_64-linux runner won't go well. But that's a problem to solve when checks exist and with knowledge of what they are (likely by narrowing systems), not by pre-emptively disabling the mechanism.

Agreed on the coverage note in your summary, and worth restating for the record: "all checks passed!" from this job means `packages` + `devShells` + `overlays`. The machines are covered by `eval-machines` and `build-pkgs`, not here.

---

PR #17, opened 2026-08-28
[launcher] Add ChatGPT and Claude desktop app launchers

cprussin:
Adds `chatgpt` and `claude` launcher apps, plus the lyra datasets that keep their local data.

- **No Linux desktop builds exist** for either: nixpkgs' `chatgpt` is the macOS `.dmg` (`platforms = darwin`) and there is no `claude-desktop` package in 26.05 or unstable. So both run as Brave `--app=` windows — chromeless, one window per site.
- New `utils/mkDesktopApp.nix`, alongside `mkWebApp`/`mkGoogleApp`, differing in that it gives each app its own `--user-data-dir` (`~/.config/<name>`) instead of opening in the main Brave profile.
- `tank/persisted-state/chatgpt` and `tank/persisted-state/claude` on lyra, so those profiles survive the tmpfs root.
- No sway change needed — the existing `^brave-.*` app_id rule matches these windows.

Not verified with `nix flake check` / `alejandra` (no nix in this environment); leaving that to CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_013VHH9pj1eDhtmmfNwE4ZnY)_

### DIFF
--- a/config/modules/ui/launcher/utils/mkDesktopApp.nix
+++ b/config/modules/ui/launcher/utils/mkDesktopApp.nix
@@ -1,0 +1,10 @@
+{
+  writeShellScript,
+  brave,
+}: name: url:
+writeShellScript name ''
+  exec ${brave}/bin/brave \
+    --app=${url} \
+    --user-data-dir="$HOME/.config/${name}" \
+    >/dev/null 2>&1
+''
--- a/config/machines/lyra/default.nix
+++ b/config/machines/lyra/default.nix
@@ -108,7 +108,9 @@ in {
         "tank/persisted-state/alsa".mountpoint = "/var/lib/alsa";
         "tank/persisted-state/argo".mountpoint = "/home/${config.primary-user.name}/.config/argo";
         "tank/persisted-state/bluetooth".mountpoint = "/var/lib/bluetooth";
+        "tank/persisted-state/chatgpt".mountpoint = "/home/${config.primary-user.name}/.config/chatgpt";
         "tank/persisted-state/chromium".mountpoint = "/home/${config.primary-user.name}/.config/chromium";
+        "tank/persisted-state/claude".mountpoint = "/home/${config.primary-user.name}/.config/claude";
         "tank/persisted-state/containers".mountpoint = "/home/${config.primary-user.name}/.local/share/containers";
         "tank/persisted-state/direnv-allow".mountpoint = "/home/${config.primary-user.name}/.local/share/direnv/allow";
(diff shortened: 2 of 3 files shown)


config/modules/ui/launcher/utils/mkDesktopApp.nix
```diff
@@ -0,0 +1,10 @@
+{
+  writeShellScript,
+  brave,
+}: name: url:
+writeShellScript name ''
+  exec ${brave}/bin/brave \
```
cprussin: This is the only place outside `pkgs/launcher/scripts/browse.nix` that hardcodes a browser binary — `mkWebApp` and `mkGoogleApp` both go through `${launcher}/bin/browse`. Consider adding an `--app`/`--profile` flag to `browse` instead and having this helper call it, so browser choice (and the `>/dev/null 2>&1` boilerplate) stays in one place.

Naming nit: `mkDesktopApp` reads as "a native desktop app", but it builds a site-in-a-window. `mkBrowserApp` / `mkSiteApp` would sit better next to `mkWebApp`.

---

config/modules/ui/launcher/utils/mkDesktopApp.nix
```diff
@@ -0,0 +1,10 @@
...
+    --app=${url} \
+    --user-data-dir="$HOME/.config/${name}" \
```
cprussin: Practical consequences of a dedicated `--user-data-dir` worth weighing:

- **First run:** a virgin profile triggers Brave's onboarding + default-browser check, which can pop a normal browser window alongside the app window. `--no-first-run --no-default-browser-check` is basically mandatory here.
- **No extensions:** Bitwarden (there's a sway rule for it) and any adblock/settings you've tuned don't exist in these profiles — you'll be typing passwords by hand into the sign-in flow.
- **Link handling:** out-of-scope navigations from a Chromium `--app=` window open a regular browser window *in the same profile*, so clicking a citation in Claude/ChatGPT lands you in a bare, extension-less Brave rather than your main browser.
- **Cost:** each profile is a full independent browser process tree (zygote, GPU, network service) rather than a cheap extra window on the already-running instance.

The alternative — plain `brave --app=URL` on the default profile, or `--profile-directory=` inside the existing `Brave-Browser` user-data-dir — gets you the chromeless window, reuses the session you already have in the browser, needs no new ZFS datasets, and just opens a window in the running instance. Independent sign-in is the only thing you give up; if that isn't an actual requirement (e.g. different accounts per app), the simpler form seems strictly better.

---

PR #18, opened 2026-08-28
[launcher] Add the Claude and ChatGPT desktop apps

cprussin:
Replaces #17, which took the wrong approach (browser app-mode windows). This packages the real apps. Two commits, independent of each other — the ChatGPT one can be dropped without touching the Claude one.

Neither app is in nixpkgs for Linux (`pkgs.chatgpt` is the macOS `.dmg`, darwin-only), and both now ship an official `.deb` from a vendor apt repo, so both packages unpack that `.deb`.

**Claude** (`pkgs/claude-desktop`) — recipe adapted from [poeck/claude-desktop-nix-flake](https://github.com/poeck/claude-desktop-nix-flake) (MIT), including the two paths patched inside `app.asar`, without which Cowork looks for QEMU's firmware under `/usr/share` and reports it missing. The bundled `virtiofsd` gets an existence check, since unlike the three rewrites nothing else fails if a release drops it. The launcher module also joins the `kvm` group, which Cowork needs for `/dev/vhost-vsock`.

**ChatGPT** (`pkgs/chatgpt-desktop`) — same shape, minus the QEMU patching. The pool URL layout and both per-arch hashes come from the AUR `chatgpt-desktop` PKGBUILD, which repackages the same binaries; the `.deb` itself is fetched from OpenAI over TLS, so a wrong third-party hash can only fail the build, never substitute content. `bin/chatgpt` is upstream's `codex-launcher`, so `~/.config/chatgpt-flags.conf` keeps working; the wrapper goes on the binary it execs. Electron's Qt theme shims and the musl/Android node prebuilds are ignored by name, so a library that does matter still fails the build.

**Datasets on lyra** (the root is a tmpfs), named for which half of each app they hold rather than for their directories, so no two differ only in case:

| dataset | mountpoint |
|---|---|
| `claude-desktop` | `~/.config/Claude` |
| `claude-code` | `~/.claude` |
| `chatgpt-desktop` | `~/.config/ChatGPT` |
| `codex` | `~/.codex` |

Both packages join the `build-pkgs` matrix — they are the derivations here whose install phases depend on content the flake doesn't pin, which is what that job exists to catch. It paid for itself immediately: the ChatGPT `.deb` ships no `usr/share/icons`, and carries libraries nothing here loads. Eval caught neither.

Verified by CI building both for real: the hashes, the asar patches, the virtiofsd and `codex-launcher` guards. Not verified — anything needing the apps running: Cowork's VM and whether its disk images want persisting too. `~/.codex` rests on Gentoo's post-install note for this package and on numtide's packaging of the same series, not on observation.

### DIFF
--- a/config/machines/lyra/default.nix
+++ b/config/machines/lyra/default.nix
@@ -108,9 +108,11 @@ in {
         "tank/persisted-state/alsa".mountpoint = "/var/lib/alsa";
         "tank/persisted-state/argo".mountpoint = "/home/${config.primary-user.name}/.config/argo";
         "tank/persisted-state/bluetooth".mountpoint = "/var/lib/bluetooth";
+        "tank/persisted-state/chatgpt-desktop".mountpoint = "/home/${config.primary-user.name}/.config/ChatGPT";
         "tank/persisted-state/chromium".mountpoint = "/home/${config.primary-user.name}/.config/chromium";
         "tank/persisted-state/claude-code".mountpoint = "/home/${config.primary-user.name}/.claude";
         "tank/persisted-state/claude-desktop".mountpoint = "/home/${config.primary-user.name}/.config/Claude";
+        "tank/persisted-state/codex".mountpoint = "/home/${config.primary-user.name}/.codex";
         "tank/persisted-state/containers".mountpoint = "/home/${config.primary-user.name}/.local/share/containers";
         "tank/persisted-state/direnv-allow".mountpoint = "/home/${config.primary-user.name}/.local/share/direnv/allow";
         "tank/persisted-state/discord".mountpoint = "/home/${config.primary-user.name}/.config/discord";
--- a/config/machines/lyra/default.nix
+++ b/config/machines/lyra/default.nix
@@ -109,6 +109,8 @@ in {
         "tank/persisted-state/argo".mountpoint = "/home/${config.primary-user.name}/.config/argo";
         "tank/persisted-state/bluetooth".mountpoint = "/var/lib/bluetooth";
         "tank/persisted-state/chromium".mountpoint = "/home/${config.primary-user.name}/.config/chromium";
+        "tank/persisted-state/claude-code".mountpoint = "/home/${config.primary-user.name}/.claude";
+        "tank/persisted-state/claude-desktop".mountpoint = "/home/${config.primary-user.name}/.config/Claude";
         "tank/persisted-state/containers".mountpoint = "/home/${config.primary-user.name}/.local/share/containers";
         "tank/persisted-state/direnv-allow".mountpoint = "/home/${config.primary-user.name}/.local/share/direnv/allow";
         "tank/persisted-state/discord".mountpoint = "/home/${config.primary-user.name}/.config/discord";
--- a/pkgs/chatgpt-desktop/derivation.nix
+++ b/pkgs/chatgpt-desktop/derivation.nix
@@ -1,0 +1,222 @@
+# OpenAI ships the Linux app as a .deb from their own apt repo and there's no
+# nixpkgs package for it (`pkgs.chatgpt` is the macOS .dmg, darwin-only), so we
+# unpack that .deb the same way `claude-desktop` does.  The URL layout and the
+# per-arch hashes come from the AUR `chatgpt-desktop` PKGBUILD, which repackages
+# the same binaries.
+#
+# `version` and `hash` move together; the pool this reads from is listed at
+# https://persistent.oaistatic.com/codex-app-prod/linux/deb/dists/stable/main/binary-amd64/Packages
+{
+  lib,
+  stdenv,
+  fetchurl,
+  alsa-lib,
+  at-spi2-core,
+  autoPatchelfHook,
+  cairo,
+  cups,
+  dbus,
+  dpkg,
+  expat,
+  fontconfig,
+  freetype,
+  gdk-pixbuf,
+  glib,
+  graphite2,
+  gtk3,
+  libGL,
+  libdrm,
+  libgbm,
+  libnotify,
+  libpulseaudio,
+  libsecret,
+  libusb1,
+  libuuid,
+  libx11,
+  libxcb,
+  libxcomposite,
+  libxcursor,
+  libxdamage,
+  libxext,
+  libxfixes,
+  libxi,
+  libxkbcommon,
+  libxrandr,
+  libxrender,
+  libxscrnsaver,
+  libxtst,
+  makeWrapper,
+  mesa,
+  nspr,
+  nss,
+  pango,
+  systemd,
+  vulkan-loader,
+  wayland,
+  wrapGAppsHook3,
+  xdg-utils,
+}: let
+  version = "26.820.71523";
+
+  sources = {
+    x86_64-linux = {
+      debArch = "amd64";
+      hash = "sha256-Ry0D6IophX8QFbK5F12AUjoTHPG8PpAX6xqP8jTeG9o=";
+    };
+    aarch64-linux = {
+      debArch = "arm64";
+      hash = "sha256-4OyqrqaqROfkWMky9c12EueGwK6wi+2XHLlq48QmYBM=";
+    };
+  };
+
+  source =
+    sources.${stdenv.hostPlatform.system}
+    or (throw "chatgpt-desktop is not packaged for ${stdenv.hostPlatform.system}");
+
+  runtimeLibs = [
+    alsa-lib
+    at-spi2-core
+    cairo
+    cups
+    dbus
+    expat
+    fontconfig
+    freetype
+    gdk-pixbuf
+    glib
+    graphite2
+    gtk3
+    libGL
+    libdrm
+    libgbm
+    libnotify
+    libpulseaudio
+    libsecret
+    libusb1
+    libuuid
+    libx11
+    libxcb
+    libxcomposite
+    libxcursor
+    libxdamage
+    libxext
+    libxfixes
+    libxi
+    libxkbcommon
+    libxrandr
+    libxrender
+    libxscrnsaver
+    libxtst
+    mesa
+    nspr
+    nss
+    pango
+    stdenv.cc.cc.lib
+    systemd
+    vulkan-loader
+    wayland
+  ];
+in
+  stdenv.mkDerivation {
+    pname = "chatgpt-desktop";
+    inherit version;
+
+    src = fetchurl {
+      url = "https://persistent.oaistatic.com/codex-app-prod/linux/deb/pool/main/c/chatgpt/chatgpt_${version}_${source.debArch}.deb";
+      inherit (source) hash;
+    };
+
+    nativeBuildInputs = [
+      autoPatchelfHook
+      dpkg
+      makeWrapper
+      wrapGAppsHook3
+    ];
+
+    buildInputs = runtimeLibs;
+
+    # Libraries the .deb ships that nothing here will load: the Qt shims
+    # Electron picks at runtime to match a Qt desktop's theme (this one is
+    # GTK), and the musl and Android builds of a few node addons, which sit
+    # beside the glibc ones that actually get loaded.  Listed out rather than
+    # blanket-ignored, so a library that does matter still fails the build.
+    autoPatchelfIgnoreMissingDeps = [
+      "libQt5Core.so.5"
+      "libQt5Gui.so.5"
+      "libQt5Widgets.so.5"
+      "libQt6Core.so.6"
+      "libQt6Gui.so.6"
+      "libQt6Widgets.so.6"
+      "libc++_shared.so"
+      "libc.musl-aarch64.so.1"
+      "libc.musl-x86_64.so.1"
+      "liblog.so"
+    ];
+
+    dontConfigure = true;
+    dontBuild = true;
+    dontStrip = true;
+    dontWrapGApps = true;
+
+    unpackPhase = ''
+      runHook preUnpack
+      dpkg-deb --fsys-tarfile "$src" | tar --extract --file - --no-same-permissions
+      runHook postUnpack
+    '';
+
+    installPhase = ''
+      runHook preInstall
+
+      mkdir -p "$out/bin" "$out/lib" "$out/share"
+      cp -a usr/lib/chatgpt "$out/lib/"
+      cp -a usr/share/applications usr/share/doc "$out/share/"
+
+      # This .deb ships no icon theme dir today, and where it would put one if
+      # that changes isn't worth guessing at.
+      for dir in icons pixmaps
+      do
+        if test -d "usr/share/$dir"
+        then
+          cp -a "usr/share/$dir" "$out/share/"
+        fi
+      done
+
+      for desktop in "$out"/share/applications/*.desktop
+      do
+        substituteInPlace "$desktop" \
+          --replace-fail "Exec=chatgpt" "Exec=$out/bin/chatgpt"
+      done
+
+      # `codex-launcher` is upstream's entry point, not the ChatGPT binary
+      # beside it: it reads ~/.config/chatgpt-flags.conf and prepends those
+      # flags before exec'ing its sibling.  It resolves that sibling from its
+      # own path, so wrapping the sibling in place below is what gets the
+      # environment onto it.
+      test -e "$out/lib/chatgpt/codex-launcher" \
+        || (echo "no codex-launcher in this .deb -- wrap ChatGPT directly" >&2; exit 1)
+      ln -s ../lib/chatgpt/codex-launcher "$out/bin/chatgpt"
+
+      runHook postInstall
+    '';
+
+    preFixup = ''
+      gappsWrapperArgs+=(
+        --prefix PATH : ${lib.makeBinPath [xdg-utils]}
+        --prefix LD_LIBRARY_PATH : ${lib.makeLibraryPath runtimeLibs}
+        --set-default ELECTRON_OZONE_PLATFORM_HINT auto
+      )
+    '';
+
+    postFixup = ''
+      wrapProgram "$out/lib/chatgpt/ChatGPT" "''${gappsWrapperArgs[@]}"
+    '';
+
+    meta = {
+      description = "Official ChatGPT desktop app for Linux";
+      homepage = "https://chatgpt.com/download";
+      license = lib.licenses.unfree;
+      mainProgram = "chatgpt";
+      platforms = builtins.attrNames sources;
+      sourceProvenance = [lib.sourceTypes.binaryNativeCode];
+    };
+  }
--- a/pkgs/claude-desktop/default.nix
+++ b/pkgs/claude-desktop/default.nix
@@ -1,0 +1,8 @@
+{nixpkgs ? (import ../../sources.nix).nixpkgs}: let
+  pkgs = import nixpkgs {
+    overlays = [
+      (import ./overlay.nix)
+    ];
+  };
+in
+  pkgs.claude-desktop
--- a/pkgs/claude-desktop/derivation.nix
+++ b/pkgs/claude-desktop/derivation.nix
@@ -1,0 +1,228 @@
+# Anthropic ships the Linux app as a .deb from their own apt repo and there's
+# no nixpkgs package for it, so we unpack that .deb ourselves.  The recipe --
+# in particular the two paths patched inside app.asar, without which Cowork
+# can't find QEMU's firmware -- is adapted from
+# https://github.com/poeck/claude-desktop-nix-flake (MIT).
+#
+# `version` and `hash` move together: bump both from the pool listing at
+# https://downloads.claude.ai/claude-desktop/apt/stable/dists/stable/main/binary-amd64/Packages
+{
+  lib,
+  stdenv,
+  fetchurl,
+  alsa-lib,
+  asar,
+  at-spi2-core,
+  autoPatchelfHook,
+  cairo,
+  cups,
+  dbus,
+  dpkg,
+  expat,
+  fontconfig,
+  freetype,
+  gdk-pixbuf,
+  glib,
+  gtk3,
+  libGL,
+  libayatana-appindicator,
+  libcap_ng,
+  libdrm,
+  libgbm,
+  libnotify,
+  libpulseaudio,
+  libseccomp,
+  libsecret,
+  libuuid,
+  libva,
+  libx11,
+  libxcb,
+  libxcomposite,
+  libxcursor,
+  libxdamage,
+  libxext,
+  libxfixes,
+  libxi,
+  libxkbcommon,
+  libxrandr,
+  libxrender,
+  libxscrnsaver,
+  libxtst,
+  makeWrapper,
+  mesa,
+  nspr,
+  nss,
+  OVMF,
+  pango,
+  perl,
+  qemu,
+  systemd,
+  trash-cli,
+  vulkan-loader,
+  wayland,
+  wrapGAppsHook3,
+  xdg-utils,
+}: let
+  version = "1.18286.2";
+
+  sources = {
+    x86_64-linux = {
+      debArch = "amd64";
+      hash = "sha256-Vvpd4FPgpo3HWDZ3hXvtz0IZsZ2QIBQA4CN7fXTVEvE=";
+    };
+    aarch64-linux = {
+      debArch = "arm64";
+      hash = "sha256-OMZaEibczHWmskGLnUwGT0+dxTMfiWCK7dVU2H1Sm6M=";
+    };
+  };
+
+  source =
+    sources.${stdenv.hostPlatform.system}
+    or (throw "claude-desktop is not packaged for ${stdenv.hostPlatform.system}");
+
+  firmwareCodePath =
+    if stdenv.hostPlatform.isAarch64
+    then "${qemu}/share/qemu/edk2-aarch64-code.fd"
+    else "${OVMF.fd}/FV/OVMF_CODE.fd";
+
+  runtimeLibs = [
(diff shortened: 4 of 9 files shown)


config/machines/lyra/default.nix
```diff
@@ -98,6 +98,7 @@ in {
         "tank/nix".mountpoint = "/nix";
         "tank/persisted-state/BitwigStudio".mountpoint = "/home/${config.primary-user.name}/.BitwigStudio";
         "tank/persisted-state/Brave-Browser".mountpoint = "/home/${config.primary-user.name}/.config/BraveSoftware/Brave-Browser";
+        "tank/persisted-state/Claude".mountpoint = "/home/${config.primary-user.name}/.config/Claude";
```
cprussin: The commit message motivates the `app.asar` patching by Cowork, but Cowork will still refuse to start on `lyra`: Anthropic's own docs ([desktop-linux, "Cowork requirements"](https://code.claude.com/docs/en/desktop-linux)) say the user must be in the `kvm` group, because Cowork also opens `/dev/vhost-vsock`, which only `kvm` group members can open — the doc explicitly says to join the group even where `/dev/kvm` is already world-accessible. The failure mode is the "Claude doesn't have permission to use virtualization (/dev/kvm)" banner.

Nothing in the tree adds it — the only `primary-user.extraGroups` entries are `dialout`, `video`, `adbusers`, `cdrom`, and `libvirtd` (crux only). Worth adding `primary-user.extraGroups = ["kvm"];` alongside the overlay in `config/modules/ui/launcher/default.nix`, or on `lyra` next to this dataset.

---

pkgs/claude-desktop/derivation.nix
```diff
@@ -0,0 +1,221 @@
...
+      asarRoot="$(mktemp -d)"
+      asar extract "$out/lib/claude-desktop/resources/app.asar" "$asarRoot"
+
+      FIRMWARE_CODE_PATH="${firmwareCodePath}" \
+      VIRTIOFSD_PATH="$out/lib/claude-desktop/resources/virtiofsd" \
+      perl -0pi -e '
+        s{([A-Za-z0-9_\$]+)=process\.arch==="arm64"\?\["/usr/share/AAVMF/AAVMF_CODE\.fd"\]:\["/usr/share/OVMF/OVMF_CODE_4M\.fd","/usr/share/OVMF/OVMF_CODE\.fd"\]}{$1=["$ENV{FIRMWARE_CODE_PATH}"]} or die "failed to patch firmware path\n";
+        s{([A-Za-z0-9_\$]+)=\["/usr/libexec/virtiofsd","/usr/bin/virtiofsd"\]}{$1=["$ENV{VIRTIOFSD_PATH}"]} or die "failed to patch virtiofsd path\n";
+        s{return A\.replace\("OVMF_CODE","OVMF_VARS"\)\.replace\("AAVMF_CODE","AAVMF_VARS"\)}{return A.replace("OVMF_CODE","OVMF_VARS").replace("AAVMF_CODE","AAVMF_VARS").replace("edk2-aarch64-code.fd","edk2-arm-vars.fd")} or die "failed to patch firmware vars path\n";
+      ' "$asarRoot/.vite/build/index.js"
```
cprussin, line 198: These three `or die`s are the whole point of adding `claude-desktop` to the `build-pkgs` matrix in `.github/workflows/ci.yml`. That job's own comment says why it exists: "a broken install phase or a stale `src` hash evaluates clean and first breaks on deploy" — and this is by far the most breakage-prone derivation in `pkgs/`, since it minifier-matches into a vendored `app.asar` and pins a `.deb` that Anthropic's pool rotates out. `eval-machines` catches neither.

Tradeoff worth stating either way: the closure pulls `qemu` + `OVMF` (both substitutable) plus a few hundred MB of `.deb`, so it's heavier than the current matrix entries but nowhere near the `emacs`/`launcher` class that the comment excludes. If you deliberately left it out, a line in that comment saying so would keep the rationale complete.

---

pkgs/claude-desktop/default.nix
```diff
@@ -0,0 +1,8 @@
+{nixpkgs ? (import ../../sources.nix).nixpkgs}: let
+  pkgs = import nixpkgs {
```
cprussin, line 2: Minor / non-blocking: this copies the `pkgs/*/default.nix` boilerplate exactly, which is the right call for consistency — but unlike every other package here, `claude-desktop` is `lib.licenses.unfree`, so this `import nixpkgs` would need `config.allowUnfree = true;` to ever produce a derivation. (Moot today: `sources.nix` isn't tracked in the repo, so none of these entry points evaluate. Just flagging that restoring it wouldn't be enough for this one.)

---

cprussin (commented):
Reviewed the derivation against upstream and against the pinned nixpkgs. The packaging itself looks correct; two things worth acting on, one nit.

**Verified good**

- **Every one of the 54 `callPackage` args resolves in the pinned `nixos-26.05` (`d407951`).** I checked the risky ones individually against that rev rather than assuming: the lowercase X11 names (`libx11`, `libxcb`, `libxcomposite`, `libxcursor`, `libxdamage`, `libxext`, `libxfixes`, `libxi`, `libxrandr`, `libxrender`, `libxscrnsaver`, `libxtst`) are real `pkgs/by-name/li/*` packages there, not `xorg.*`-only attrs — upstream builds on `nixos-unstable`, so this was the obvious place for a 26.05 break, and there isn't one. `libgbm`, `mesa`, `libGL` and `OVMF` come from `all-packages.nix`; `libxkbcommon`, `fontconfig` and `makeWrapper` likewise. `${OVMF.fd}/FV/OVMF_CODE.fd` matches the `fd` output layout at that rev (`fwPrefix = "OVMF"` on x86_64).
- **Faithful to upstream.** `diff` of the phase bodies against `poeck/claude-desktop-nix-flake@main` is byte-identical modulo alejandra's indent — all three perl regexes, both env vars, the `--unpack "*.node"`, the wrapper args. `version` and both per-arch hashes match. The only deliberate deltas are dropping `finalAttrs` (fine, `version` moved to the `let`), `passthru.updateScript`/`sources`, and `meta.maintainers` — all correct to drop for a vendored copy.
- **Conventions.** `default.nix`/`derivation.nix`/`overlay.nix` split matches `powerpanel`/`launcher`; `self: _:` overlay form matches; registering in the launcher module rather than `system/nix` follows the precedent already set by `pkgs/launcher/overlay.nix` two lines below. Dataset name and alphabetical position in `mkZfsFileSystems` are right, and `~/.config/Claude` is the correct Electron `userData` dir for `productName = "Claude"`. `deadnix` should be clean — I checked all 54 args are referenced.
- **Runtime.** `--set-default ELECTRON_OZONE_PLATFORM_HINT auto` is equivalent to the `--ozone-platform-hint=auto` the other Electron entries pass, and no sway rule is needed (none of `slack`/`discord`/`element` have one either). `dontWrapGApps` + appending to `gappsWrapperArgs` in `preFixup` works because `runHook preFixup` runs `preFixupHooks` (where `gappsWrapperArgsHook` lives) before the `preFixup` attribute.

**Worth acting on**

1. `lyra` is missing the `kvm` group — Cowork, which is the stated reason for the firmware patching, won't run without it. See the inline comment.
2. `claude-desktop` isn't in the `build-pkgs` CI matrix, and it's the one package in `pkgs/` whose install phase can fail on content the flake doesn't pin. See the inline comment.

---

config/machines/lyra/default.nix
cprussin: Right — fixed in 548537e. Put it in the launcher module rather than on `lyra`, next to the overlay, so it travels with the app rather than with one machine's disk layout, with the `/dev/vhost-vsock` reason in a comment.

---

pkgs/claude-desktop/derivation.nix
cprussin: Added to the matrix in 548537e — not a deliberate omission, I'd only skimmed the first two jobs when I wrote this. It's the more valuable half of the fix: nothing in this session can build the derivation, so CI is now the first thing that actually exercises the `.deb` hash and the three `or die`s.

---

pkgs/claude-desktop/default.nix
cprussin: Agreed, and leaving it as-is: these entry points are already dead across every package (no `sources.nix` in the tree), so making this one the sole file that differs would trade a real inconsistency for a hypothetical fix. Whoever restores `sources.nix` can add `config.allowUnfree` here in the same pass.

---

config/machines/lyra/default.nix
cprussin: `~/.claude` is not persisted, and `/` is tmpfs here.

`~/.config/Claude` is the Electron `userData` dir (login session, window state). But the app's whole point in this PR is the integrated Claude Code, which keeps its own state in `~/.claude` — `settings.json`, `.credentials.json`, `projects/` session history, MCP server config, plugins. All of that is lost on every reboot as written.

Suggest a second dataset next to this one:

```nix
"tank/persisted-state/claude".mountpoint = "/home/${config.primary-user.name}/.claude";
```

(Also worth checking after first run whether Cowork's VM disk images land under `~/.config/Claude` or somewhere in `~/.local/share` / `~/.cache` — if the latter, re-downloading the guest image every boot would be unpleasant.)

---

pkgs/claude-desktop/derivation.nix
```diff
@@ -0,0 +1,221 @@
...
+      cp -a usr/share/applications usr/share/icons usr/share/doc "$out/share/"
+
+      substituteInPlace "$out/share/applications/claude-desktop.desktop" \
+        --replace-fail "Exec=claude-desktop" "Exec=$out/bin/claude-desktop"
+
+      asarRoot="$(mktemp -d)"
+      asar extract "$out/lib/claude-desktop/resources/app.asar" "$asarRoot"
+
+      FIRMWARE_CODE_PATH="${firmwareCodePath}" \
+      VIRTIOFSD_PATH="$out/lib/claude-desktop/resources/virtiofsd" \
```
cprussin, line 193: The three JS rewrites below each have an `or die`, so a .deb whose bundle no longer matches fails the build loudly — good. But `VIRTIOFSD_PATH` is only interpolated: nothing checks that `resources/virtiofsd` actually exists in the .deb.

Anthropic's docs describe the bundled copy as a fallback for distros without a `virtiofsd` package ("Ubuntu 22.04 has no `virtiofsd` package; the app uses a bundled copy there"), so it's plausible it gets dropped in a future release once that's no longer needed. If it does, this builds clean and Cowork breaks at runtime with a path into the store that doesn't exist.

Cheap guard in the same spirit as the `or die`s:

```sh
test -x "$out/lib/claude-desktop/resources/virtiofsd" \
  || (echo "bundled virtiofsd is gone; point at pkgs.virtiofsd instead" >&2; exit 1)
```

nixpkgs has a `virtiofsd` package, so the fallback is a one-word change when that day comes.

---

pkgs/claude-desktop/derivation.nix
```diff
@@ -0,0 +1,221 @@
...
+      rm "$out/lib/claude-desktop/resources/app.asar"
+      asar pack --unpack "*.node" "$asarRoot" "$out/lib/claude-desktop/resources/app.asar"
+
+      runHook postInstall
+    '';
+
+    preFixup = ''
+      gappsWrapperArgs+=(
+        --prefix PATH : ${lib.makeBinPath runtimeBins}
+        --prefix LD_LIBRARY_PATH : ${lib.makeLibraryPath runtimeLibs}
```
cprussin, line 209: Nit, only if you hit trouble later: this `LD_LIBRARY_PATH` is exported into the app's environment and therefore inherited by everything it spawns — including `qemu-system-x86_64` and `virtiofsd`, which are correctly linked already. Prefixing ~40 Electron/GTK libs (mesa, libGL, vulkan-loader, nss...) onto their search path is a plausible source of odd Cowork-only failures.

`autoPatchelfHook` already fixes RPATHs for everything under `$out`, so this is really only covering Electron's runtime `dlopen`s. Worth remembering as the first thing to try if QEMU misbehaves.

---

cprussin (commented):
Second-round review. Both first-round fixes check out; one real gap found (`~/.claude` not persisted) plus two small packaging notes.

## The two fixes

**`primary-user.extraGroups = ["kvm"]` — correct, and sufficient.** Verified against the pinned nixpkgs (`d407951`):

- `users.groups.kvm` is defined unconditionally in `nixos/modules/config/users-groups.nix` (`kvm.gid = ids.gids.kvm;`), so no missing-group assertion.
- `primary-user.extraGroups` is an alias to `users.users.<name>.extraGroups` (list-merged with the `wheel` default), so this composes fine.
- The comment's claim holds: systemd's `50-udev-default.rules` gives `/dev/vhost-vsock` `GROUP="kvm" MODE="0660"` while `/dev/kvm` is `0666`.
- I checked whether `boot.kernelModules = ["vhost_vsock"]` is also needed, since the linked docs call out `modprobe vhost_vsock` as a separate failure mode. It isn't: `vhost_vsock` carries `MODULE_ALIAS("devname:vhost-vsock")`, `kmod-static-nodes.service` is in `additionalUpstreamSystemUnits`, and `systemd.nix` overrides its `ConditionFileNotEmpty` to `/run/booted-system/kernel-modules/lib/modules/%v/modules.devname`. So the node exists at boot with the right group and opening it autoloads the module. lyra already has `kvm-amd`.

Placement in the launcher module rather than `modules/system/devices/` reads fine to me — `machines/crux/libvirtd.nix` sets the group next to the thing that needs it, and lyra is the only machine importing `profiles/pc` today.

**CI matrix entry — correct.** `claude-desktop` matches the overlay attribute (the matrix keys off attr names, not directories — cf. `pass`/`sudo` vs `pass-with-otp`/`sudo-with-insults`), and it's in alphabetical position. Unfree is a non-issue: `allowUnfree` comes from `config/modules/system/nix/nixpkgs-config.nix` via `nixpkgs.config`, which `eval-machines` already exercises on `discord`/`slack`/`steam`. Confirmed `OVMF` has an `fd` output containing `FV/OVMF_{CODE,VARS}.fd`, so the interpolation and the `OVMF_CODE`→`OVMF_VARS` rewrite both resolve. On timeout: the job will substitute a few GB (qemu, OVMF, mesa, the Electron stack) rather than build it, so it isn't in the same class as the `emacs`/`launcher` exclusions the comment above the matrix describes — worth revisiting that comment only if the job actually starts crowding 30 minutes.

## Findings

1. `~/.claude` isn't persisted — see the `lyra/default.nix` comment. This is the one I'd actually fix before merging.
2. Missing existence check on the bundled `virtiofsd` — see the `derivation.nix` comment.
3. `LD_LIBRARY_PATH` leaking into spawned QEMU — noted for future debugging, no action needed.

Everything else in the diff (overlay wiring, alphabetical insertions, `dontWrapGApps` + manual `makeWrapper` ordering, `preFixup`/`postFixup` vs `autoPatchelf` hook ordering, arg names against the pinned rev) looks right.

---

config/machines/lyra/default.nix
cprussin: Good catch — added in 9bcd573. The two datasets end up differing only in case (`Claude` for `~/.config/Claude`, `claude` for `~/.claude`), which is a little unfortunate to read in `zfs list`, but each is named for the directory it holds like every other entry, so I kept the convention rather than inventing a suffix.

Cowork's VM image location I can't check from here — no way to run the app — so leaving that for the first launch.

---

pkgs/claude-desktop/derivation.nix
cprussin: Added in 9bcd573, with `test -e` rather than `-x`: the unpack passes `--no-same-permissions`, so exec bits are subject to the build umask and a mode check could fail for the wrong reason. Presence is the thing the interpolation actually depends on.

---

pkgs/claude-desktop/derivation.nix
cprussin: Noted, no change — agreed it's the first thing to try if Cowork misbehaves. Narrowing it (dropping `LD_LIBRARY_PATH` and letting `autoPatchelfHook` cover the RPATHs, or scrubbing it in a QEMU wrapper) is a change I'd only want to make against a machine that can actually run the app, since the failure mode is a `dlopen` at runtime that no build catches.

---

cprussin (commented):
Round 3 — reviewed only `git diff 548537e..HEAD` (the virtiofsd guard + the `tank/persisted-state/claude` dataset). Delta is clean, no findings.

Verified:

- **Guard shell semantics.** `test -e … || (echo …; exit 1)` aborts the build correctly. The subshell's `exit 1` makes the `||` list return 1, and since the list is the whole command (not the LHS of another `&&`/`||`, not in a condition context), errexit fires. Confirmed under nixpkgs stdenv's exact options (`set -eu -o pipefail` + `shopt -s inherit_errexit`): script stops, message on stderr, exit 1. An `if ! …; then` form would exit without relying on errexit, but that's style, not a bug.
- **Placement.** After `cp -a usr/lib/claude-desktop "$out/lib/"` (so the path exists to test) and before `asar extract` / the perl patch — fails fast, before the expensive unpack, and the tested path is byte-identical to the one interpolated into `VIRTIOFSD_PATH`.
- **`-e` vs `-x`.** Right call given `tar --no-same-permissions` in `unpackPhase`; `-e` can't false-fail on a mode difference, and a missing binary is the failure mode the check exists for.
- **Message.** Actionable and names the fix (`pkgs.virtiofsd`); the comment above it correctly explains why this one needs a guard while the three rewrites carry their own `or die`.
- **Dataset.** ASCII-sorted list (uppercase block first); `claude` sits correctly between `chromium` and `containers`, and is distinct from the `Claude` → `~/.config/Claude` entry in the uppercase block. Nothing in the repo writes `~/.claude` declaratively, so no conflict with the mountpoint.
- **Commit message.** Matches the diff — the virtiofsd-check paragraph and the "two datasets on lyra" paragraph both describe what's actually here.

LGTM (posting as a comment rather than an approval — GitHub won't let me approve my own PR).

---

pkgs/chatgpt-desktop/derivation.nix
```diff
@@ -0,0 +1,212 @@
...
+    preFixup = ''
+      gappsWrapperArgs+=(
+        --prefix PATH : ${lib.makeBinPath [xdg-utils]}
+        --prefix LD_LIBRARY_PATH : ${lib.makeLibraryPath runtimeLibs}
+        --set-default ELECTRON_OZONE_PLATFORM_HINT auto
+      )
+    '';
+
+    postFixup = ''
+      makeWrapper "$out/lib/chatgpt/ChatGPT" "$out/bin/chatgpt" \
```
cprussin: Wrapping `ChatGPT` skips the launcher the .deb ships.

The .deb also ships `usr/lib/chatgpt/codex-launcher`, and that is what upstream's own entry point is: the Gentoo ebuild does `dosym -r /opt/chatgpt-desktop/codex-launcher /usr/bin/chatgpt`, and both existing Nix packagings do the same (`numtide/llm-agents.nix` symlinks `bin/chatgpt` -> `lib/chatgpt/codex-launcher`; `joshsymonds/nix-config` does `makeWrapper "$out/lib/chatgpt/codex-launcher" "$out/bin/chatgpt"`). The AUR replaces that launcher with its own `chatgpt-launcher.sh`, which shows what it is for: reading `${XDG_CONFIG_HOME:-~/.config}/chatgpt-flags.conf` and prepending those flags before `exec`ing the sibling `ChatGPT`.

So as written, `chatgpt-flags.conf` (and anything else upstream's launcher does — it may also default ozone/password-store flags) is silently dropped. Wrapping `codex-launcher` instead is safe here: it resolves its sibling via `readlink -f "$0"`, and makeWrapper execs the absolute store path, so `$0` still lands in `$out/lib/chatgpt`.

---

pkgs/chatgpt-desktop/derivation.nix
```diff
@@ -0,0 +1,212 @@
...
+    # they're an alternative to.  Listed out rather than blanket-ignored, so a
+    # library that actually matters still fails the build.
+    autoPatchelfIgnoreMissingDeps = [
+      "libQt5Core.so.5"
+      "libQt5Gui.so.5"
+      "libQt5Widgets.so.5"
+      "libQt6Core.so.6"
+      "libQt6Gui.so.6"
+      "libQt6Widgets.so.6"
+      "libc.musl-x86_64.so.1"
```
cprussin, line 152: Two gaps in this list:

1. `libc.musl-x86_64.so.1` is the x86_64 soname only, but `meta.platforms` claims `aarch64-linux`, where the same addons carry `libc.musl-aarch64.so.1`. An aarch64 build would fail here.
2. `numtide/llm-agents.nix`, packaging this same 26.820.x series, also needs `libc++_shared.so` and `liblog.so` — the Android prebuilds that sit next to the musl/glibc ones in those addons' `prebuilds/` dirs.

Alternative that avoids guessing at sonames entirely (what the Gentoo ebuild and `joshsymonds/nix-config` both do): drop the non-native prebuilds before autoPatchelf runs — `find "$out/lib/chatgpt" -type d -name prebuilds` then delete every child but `linux-x64`/`linux-arm64`, plus `-name '*.musl.node' -delete`. That also trims the closure. The Qt shim entries are still needed either way.

---

config/machines/lyra/default.nix
```diff
@@ -98,6 +98,8 @@ in {
         "tank/nix".mountpoint = "/nix";
         "tank/persisted-state/BitwigStudio".mountpoint = "/home/${config.primary-user.name}/.BitwigStudio";
         "tank/persisted-state/Brave-Browser".mountpoint = "/home/${config.primary-user.name}/.config/BraveSoftware/Brave-Browser";
+        "tank/persisted-state/ChatGPT".mountpoint = "/home/${config.primary-user.name}/.config/ChatGPT";
```
cprussin: `~/.config/ChatGPT` is persisted, but `~/.codex` (dropped in the amend) isn't — and `/` is tmpfs here, so it goes away every boot.

The app is the Codex app (`codex-app-prod`), and Gentoo's `pkg_postinst` for this exact package says: "~/.codex is shared with dev-util/codex, so signing in may update its login state." So Codex auth/config written there is lost on reboot. Was removing the dataset deliberate? If it was dropped only because nothing creates the dir, note the ZFS mount creates it at boot, which is harmless.

---

pkgs/chatgpt-desktop/derivation.nix
```diff
@@ -0,0 +1,212 @@
...
+      # that changes isn't worth guessing at.
+      for dir in icons pixmaps
+      do
+        if test -d "usr/share/$dir"
+        then
+          cp -a "usr/share/$dir" "$out/share/"
+        fi
+      done
+
+      # The .deb's own Exec= points at the /usr path its launcher lives at, so
```
cprussin: Nit, comment is inaccurate: upstream's `chatgpt.desktop` has a bare `Exec=chatgpt %U`, not a `/usr` path (the Gentoo ebuild seds `s|^Exec=chatgpt|...|`, and `joshsymonds` rewrites to `Exec=chatgpt %U`). The `[^ ]*` rewrite still does the right thing; only the justification is wrong.

Two small things while here: unlike `claude-desktop`'s `--replace-fail`, this silently no-ops if a future .desktop has no `Exec=` line; and `usr/share/doc/chatgpt/copyright` (the unfree license text, which `claude-desktop` keeps) isn't copied.

---

cprussin (commented):
Reviewed the ChatGPT commit only (`9bcd573..2d24067`); the claude-desktop commit is untouched.

**Provenance checks out.** Against the AUR `chatgpt-desktop` PKGBUILD/.SRCINFO:
- version `26.820.71523` matches;
- both hashes are byte-identical transcriptions — `472d03e8…de1bda` -> `sha256-Ry0D6IophX8QFbK5F12AUjoTHPG8PpAX6xqP8jTeG9o=` (amd64) and `e0ecaaae…266013` -> `sha256-4OyqrqaqROfkWMky9c12EueGwK6wi+2XHLlq48QmYBM=` (arm64);
- the URL pattern matches `source_x86_64`/`source_aarch64` exactly;
- the binary really is at `usr/lib/chatgpt/ChatGPT` (Gentoo `pax-mark`s it; the shipped AppArmor profile names that path).

**Deps look complete.** Everything in the PKGBUILD `depends` list maps to something in `buildInputs`/PATH (`libcups`->cups, `libglvnd`->libGL, `libgcc`/`libstdc++`->`stdenv.cc.cc.lib`, `systemd-libs`->systemd, `vulkan-driver`->vulkan-loader, `libusb`->libusb1), plus `libgbm` split out for 25.11+, which Arch gets via mesa. No unused `callPackage` args (deadnix clean), and every arg is a real top-level nixpkgs attr — the set is claude-desktop's plus `graphite2` and `libusb1`. The icons fix in the amend is right: this .deb ships `usr/share/pixmaps/chatgpt.png`, not an icon theme dir (confirmed by both the Gentoo ebuild and `numtide/llm-agents.nix`), so the conditional loop picks it up.

Findings inline; the two worth acting on are the wrapper target (`ChatGPT` vs the `codex-launcher` the .deb ships) and the incomplete `autoPatchelfIgnoreMissingDeps` list (aarch64 musl soname; the Android prebuild sonames `numtide` hits on this same version series). Note CI hadn't finished the `build pkgs.chatgpt-desktop` job when I wrote this, so the autopatchelf point is from comparing against other packagings, not an observed failure.

Two minor, non-blocking notes not worth inline comments:
- `pipewire` isn't in `runtimeLibs`; the AUR lists it as an optdepend for WebRTC screen sharing under Wayland, and `numtide` carries it in `runtimeDependencies`. Same story for `libayatana-appindicator` (tray) which `claude-desktop` does carry.
- Upstream is already at `26.820.80927`; the AUR pin is a couple of releases behind. Fine to land as-is.

---

PR #19, opened 2026-08-28
[launcher] Fix the ChatGPT app's store copies, environment and Wayland backend

cprussin:
Follow-up to #18. Three defects, two commits.

### 1. EACCES under `~/.codex` on every launch

The app materializes its bundled plugins there with node's `fs.cp`, which preserves the mode of what it copies — and everything it copies comes out of the Nix store read-only. It then can't rewrite the manifests it just wrote, so `bundled_plugins_marketplace_resolve_failed` every launch.

`patch-asar.py` is vendored from [numtide/llm-agents.nix](https://github.com/numtide/llm-agents.nix) (MIT), which hit the same two NixOS-specific problems: that copy (rewritten to `cp` + `chmod u+w`, hence coreutils on PATH), and a `process.report` probe in `@parcel/watcher` that trips a CFI guard in the bundled Electron. Both abort the build if the minified code they match moves.

### 2. Sign-in fails — LD_LIBRARY_PATH leaking into every child

Symptom: a flash of "Continue to sign in on your browser", no browser, then the sign-in screen again; a second click says a login is already in progress.

Login is driven by the bundled `codex` binary, which spawns the browser itself. Run that binary from a shell and the browser comes up; run it as a child of the app and it doesn't. The wrapper was exporting an `LD_LIBRARY_PATH` of forty Electron/GTK libraries, inherited by `codex`, by `xdg-open`, and by the browser those reach.

Electron needs them only on the objects that dlopen them — autoPatchelf's `runtimeDependencies`. numtide reached the same conclusion in as many words: *"without leaking a broad LD_LIBRARY_PATH into Electron's Node and Chromium children."*

Worth recording: the round-2 reviewer on #18 flagged exactly this as "the first thing to try if QEMU misbehaves". It was a defect, not a note. **`claude-desktop` still carries the same LD_LIBRARY_PATH** — untouched here since nothing has misbehaved, but that's where to start if Cowork's QEMU ever does.

### 3. XWayland instead of native Wayland

`ELECTRON_OZONE_PLATFORM_HINT` is a hint this Electron fork ignores. Replaced with `--ozone-platform=wayland`, guarded on `WAYLAND_DISPLAY` so an X11 session is unaffected.

**Verified by CI only** — it builds, and the asar patches are proven to apply. Whether 2 and 3 do what they claim needs your machine: after deploying, sign-in should reach the browser, and `swaymsg -t get_tree | grep -i chatgpt` should show an `app_id` rather than X11 `window_properties`.

### DIFF
--- a/pkgs/chatgpt-desktop/derivation.nix
+++ b/pkgs/chatgpt-desktop/derivation.nix
@@ -14,6 +14,7 @@
   at-spi2-core,
   autoPatchelfHook,
   cairo,
+  coreutils,
   cups,
   dbus,
   dpkg,
@@ -50,6 +51,7 @@
   nspr,
   nss,
   pango,
+  python3,
   systemd,
   vulkan-loader,
   wayland,
@@ -72,6 +74,13 @@
   source =
     sources.${stdenv.hostPlatform.system}
     or (throw "chatgpt-desktop is not packaged for ${stdenv.hostPlatform.system}");
+
+  # coreutils is load-bearing: the asar patch below makes the app copy its
+  # bundled plugins with `cp` and `chmod` rather than node's fs.cp.
+  runtimeBins = [
+    coreutils
+    xdg-utils
+  ];
 
   runtimeLibs = [
     alsa-lib
@@ -130,6 +139,7 @@ in
       autoPatchelfHook
       dpkg
       makeWrapper
+      python3
       wrapGAppsHook3
     ];
 
@@ -187,6 +197,14 @@ in
           --replace-fail "Exec=chatgpt" "Exec=$out/bin/chatgpt"
       done
 
+      # Two things the app does that only break on NixOS: it materializes its
+      # bundled plugins into ~/.codex with node's fs.cp, which preserves the
+      # store's read-only modes and then can't write the manifests it rewrites
+      # (EACCES on every launch), and @parcel/watcher probes `process.report`,
+      # which trips a CFI guard in the bundled Electron runtime.  Both patches
+      # fail the build if the code they match has moved.
+      python3 ${./patch-asar.py} "$out/lib/chatgpt/resources/app.asar"
+
       # `codex-launcher` is upstream's entry point, not the ChatGPT binary
       # beside it: it reads ~/.config/chatgpt-flags.conf and prepends those
       # flags before exec'ing its sibling.  It resolves that sibling from its
@@ -201,7 +219,7 @@ in
 
     preFixup = ''
       gappsWrapperArgs+=(
-        --prefix PATH : ${lib.makeBinPath [xdg-utils]}
+        --prefix PATH : ${lib.makeBinPath runtimeBins}
         --prefix LD_LIBRARY_PATH : ${lib.makeLibraryPath runtimeLibs}
         --set-default ELECTRON_OZONE_PLATFORM_HINT auto
       )
--- a/pkgs/chatgpt-desktop/derivation.nix
+++ b/pkgs/chatgpt-desktop/derivation.nix
@@ -51,6 +51,7 @@
   nspr,
   nss,
   pango,
+  pipewire,
   python3,
   systemd,
   vulkan-loader,
@@ -145,6 +146,26 @@ in
 
     buildInputs = runtimeLibs;
 
+    # Electron dlopens these rather than linking them, so autoPatchelf has to
+    # be told: they go on each executable's RPATH.  What they must not do is
+    # arrive as an LD_LIBRARY_PATH on the app's environment -- see the wrapper
+    # below.
+    runtimeDependencies = [
+      libGL
+      libgbm
+      libnotify
+      libpulseaudio
+      libsecret
+      pipewire
+      vulkan-loader
+      wayland
+      # systemd has no `lib` output at this nixpkgs rev, so this resolves to
+      # the same path as bare `systemd` and is only here to stay correct if one
+      # is ever added.  libudev.so.1 does live in its $out/lib, which is where
+      # the hook looks: it appends /lib to whatever path it is given.
+      (lib.getLib systemd)
+    ];
+
     # Libraries the .deb ships that nothing here will load: the Qt shims
     # Electron picks at runtime to match a Qt desktop's theme (this one is
     # GTK), and the musl and Android builds of a few node addons, which sit
@@ -206,10 +227,11 @@ in
       python3 ${./patch-asar.py} "$out/lib/chatgpt/resources/app.asar"
 
       # `codex-launcher` is upstream's entry point, not the ChatGPT binary
-      # beside it: it reads ~/.config/chatgpt-flags.conf and prepends those
-      # flags before exec'ing its sibling.  It resolves that sibling from its
-      # own path, so wrapping the sibling in place below is what gets the
-      # environment onto it.
+      # beside it: a two-line script that resolves its sibling from its own
+      # path and execs it.  The wrapper goes on that sibling because ChatGPT is
+      # the process that needs the flags and the environment, and wrapping it
+      # also covers anything that runs it directly rather than via the
+      # launcher.
       test -e "$out/lib/chatgpt/codex-launcher" \
         || (echo "no codex-launcher in this .deb -- wrap ChatGPT directly" >&2; exit 1)
       ln -s ../lib/chatgpt/codex-launcher "$out/bin/chatgpt"
@@ -217,16 +239,29 @@ in
       runHook postInstall
     '';
 
+    # The ozone flag rather than ELECTRON_OZONE_PLATFORM_HINT, which this
+    # Electron fork ignores; guarded on WAYLAND_DISPLAY so an X11 session is
+    # unaffected.  Being flagsBefore, a `--ozone-platform=` passed on the
+    # command line still wins, which is the way back to XWayland.
     preFixup = ''
       gappsWrapperArgs+=(
-        --prefix PATH : ${lib.makeBinPath runtimeBins}
-        --prefix LD_LIBRARY_PATH : ${lib.makeLibraryPath runtimeLibs}
-        --set-default ELECTRON_OZONE_PLATFORM_HINT auto
+        # --suffix, not --prefix (which is what claude-desktop uses): every
+        # child inherits this, including the codex agent shelling out in a
+        # project of mine, and it should get my coreutils rather than the
+        # app's.  The store copy is only the fallback that makes the asar
+        # patch's `cp`/`chmod` work when PATH has neither.
+        --suffix PATH : ${lib.makeBinPath runtimeBins}
+        --add-flags "\''${WAYLAND_DISPLAY:+--ozone-platform=wayland}"
       )
     '';
 
+    # `wrapProgramShell`, not `wrapProgram`: wrapGAppsHook3 propagates
+    # makeBinaryWrapper, which redefines `wrapProgram` to emit a C wrapper that
+    # passes --add-flags through verbatim -- the `${WAYLAND_DISPLAY:+...}` above
+    # would reach Electron as a literal argument instead of expanding.  The
+    # shell flavour stays available because makeWrapper is in nativeBuildInputs.
     postFixup = ''
-      wrapProgram "$out/lib/chatgpt/ChatGPT" "''${gappsWrapperArgs[@]}"
+      wrapProgramShell "$out/lib/chatgpt/ChatGPT" "''${gappsWrapperArgs[@]}"
     '';
 
     meta = {
--- a/pkgs/chatgpt-desktop/patch-asar.py
+++ b/pkgs/chatgpt-desktop/patch-asar.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python3
+"""Apply byte-length-preserving source patches inside app.asar.
+
+Vendored from numtide/llm-agents.nix (MIT), which hit the same two problems.
+
+The asar header records file offsets, so every replacement is padded with
+spaces to the original's exact byte length instead of re-packing the archive.
+Minified identifiers change between releases, so patterns are regexes that
+capture the identifiers they need instead of hard-coding them.
+"""
+
+import re
+import sys
+from collections.abc import Callable
+from pathlib import Path
+
+# @parcel/watcher uses detect-libc in a named worker. Its process.report
+# fallback trips a CFI guard in the bundled Owl/Electron runtime on NixOS.
+# detect-libc falls back to its ELF/filesystem/ldd probes instead.
+SKIP_PROCESS_REPORT = (
+    re.compile(rb"isLinux\(\) && process\.report"),
+    lambda _m: b"false /* nix:skip report */",
+)
+
+# The app materializes bundled plugins in ~/.codex and rewrites selected
+# manifests there. Node's fs.cp preserves the Nix store's read-only modes,
+# so copy with coreutils and make only the user-owned destination writable.
+# `exec` is the promisified execFile helper already used for the darwin
+# `ditto` branch. Hoisting `platform` into a local buys the bytes needed.
+COPY_PLUGINS_WRITABLE = (
+    re.compile(
+        rb"(?P<fn>async function [\w$]+\(e,t\)\{)"
+        rb"if\((?P<plat>[\w$]+\.default\.platform)===`darwin`\)"
+        rb"(?P<ditto>\{await (?P<exec>[\w$]+)\(`/usr/bin/ditto`,\[`--noqtn`,e,t\]\);return\})"
+        rb"if\((?P=plat)!==`win32`\)\{"
+        rb"await [\w$]+\.default\.cp\(e,t,\{recursive:!0,verbatimSymlinks:!0\}\);return\}"
+    ),
+    lambda m: (
+        m["fn"]
+        + b"let r="
+        + m["plat"]
+        + b";if(r===`darwin`)"
+        + m["ditto"]
+        + b"if(r!==`win32`){await "
+        + m["exec"]
+        + b"(`cp`,[`-r`,e+`/.`,t]);await "
+        + m["exec"]
+        + b"(`chmod`,[`-R`,`u+w`,t]);return}"
+    ),
+)
+
+PATCHES: list[tuple[re.Pattern[bytes], Callable[[re.Match[bytes]], bytes]]] = [
+    SKIP_PROCESS_REPORT,
+    COPY_PLUGINS_WRITABLE,
+]
+
+
+def main() -> None:
+    """Patch the asar archive given as the only argument."""
+    asar = Path(sys.argv[1])
+    data = asar.read_bytes()
+    for pattern, build in PATCHES:
+        matches = list(pattern.finditer(data))
+        if len(matches) != 1:
+            sys.exit(
+                f"expected 1 match for {pattern.pattern[:60]!r} in {asar}, got {len(matches)}"
+            )
+        m = matches[0]
+        original = m.group(0)
+        replacement = build(m)
+        if len(replacement) > len(original):
+            sys.exit(f"replacement longer than original: {replacement[:60]!r}...")
+        data = (
+            data[: m.start()] + replacement.ljust(len(original), b" ") + data[m.end() :]
+        )
+    asar.write_bytes(data)
+
+
+if __name__ == "__main__":
+    main()
--- a/pkgs/chatgpt-desktop/patch-asar.py
+++ b/pkgs/chatgpt-desktop/patch-asar.py
@@ -1,7 +1,29 @@
 #!/usr/bin/env python3
 """Apply byte-length-preserving source patches inside app.asar.
 
-Vendored from numtide/llm-agents.nix (MIT), which hit the same two problems.
+Vendored from numtide/llm-agents.nix, packages/chatgpt/patch-asar.py, taken
+from main on 2026-08-28; diff against that path to re-sync when a version bump
+moves the minified code these patterns match.
+
+Copyright (c) 2024 Numtide
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the "Software"), to deal in
+the Software without restriction, including without limitation the rights to
+use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+of the Software, and to permit persons to whom the Software is furnished to do
+so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
 
 The asar header records file offsets, so every replacement is padded with
 spaces to the original's exact byte length instead of re-packing the archive.


pkgs/chatgpt-desktop/derivation.nix
```diff
@@ -130,11 +140,26 @@ in
...
+    # below.
+    runtimeDependencies = [
+      libGL
+      libgbm
+      libnotify
+      libsecret
+      pipewire
+      vulkan-loader
+      wayland
+    ];
```
cprussin, line 167: Two libraries the old `LD_LIBRARY_PATH` covered are dropped here and won't come back via autoPatchelf, because Chromium `dlopen`s them (generated stubs) rather than linking them, so they never appear in any `DT_NEEDED`:

- **`libpulseaudio`** — `media/audio/pulse` `dlopen`s `libpulse.so.0`. Without it Chromium falls back to ALSA/no audio; voice mode is the visible casualty. The AUR PKGBUILD hard-depends on `libpulse` for the same reason.
- **`libudev`** (from `systemd`) — `device/udev_linux/udev_loader` `dlopen`s `libudev.so.1`. Loses USB/HID/gamepad enumeration and device monitoring.

The comparator I'd trust most is nixpkgs' own prebuilt Electron, which sets exactly this set on the RPATH rather than an `LD_LIBRARY_PATH` (`pkgs/development/tools/electron/binary/generic.nix`, `electronLibPath`): `systemd`, `libnotify`, `pipewire`, `libsecret`, `libpulseaudio`, `libgbm`, `libGL`, `vulkan-loader` (+ `speechd`, which was never in scope here). This list is that set minus `systemd` and `libpulseaudio`.

Caveat: I can't verify empirically — no nix here, and `persistent.oaistatic.com` is egress-blocked, so I couldn't read the ELF headers. If you can run `patchelf --print-needed` on the built `ChatGPT` and both sonames are there, ignore this.

Suggested:

```nix
    runtimeDependencies = [
      libGL
      libgbm
      libnotify
      libpulseaudio
      libsecret
      pipewire
      (lib.getLib systemd)
      vulkan-loader
      wayland
    ];
```

`lib.getLib` is load-bearing for `systemd` specifically. The hook does a plain string splice on the *default* output — `--runtime-dependencies "${runtimeDependenciesArray[@]/%//lib}"` in `auto-patchelf.sh` — and systemd's `out/lib` holds units and udev rules, not `libudev.so.1`, which lives in the `lib` output. Plain `systemd` here would add a real directory that happens to contain nothing useful, i.e. silently do nothing. (The seven currently listed are all single-lib-output, so they're fine as written.)

---

pkgs/chatgpt-desktop/derivation.nix
```diff
@@ -201,9 +234,8 @@ in
 
     preFixup = ''
       gappsWrapperArgs+=(
-        --prefix PATH : ${lib.makeBinPath [xdg-utils]}
-        --prefix LD_LIBRARY_PATH : ${lib.makeLibraryPath runtimeLibs}
-        --set-default ELECTRON_OZONE_PLATFORM_HINT auto
+        --prefix PATH : ${lib.makeBinPath runtimeBins}
+        --add-flags "\''${WAYLAND_DISPLAY:+--ozone-platform=wayland}"
       )
```
cprussin, line 255: Traced this end to end against `pkgs/build-support/setup-hooks/make-wrapper.sh` at the pinned rev — **the expansion does survive**, no change needed. For the record, since it's fragile enough to be worth writing down:

- Nix indented string: a bare `\` is literal and `''$` is an escaped `$`, so `\''${...}` yields `\${...}` in the `preFixup` script.
- Bash: `"\${...}"` stores the literal `${WAYLAND_DISPLAY:+--ozone-platform=wayland}` in `gappsWrapperArgs`.
- `--add-flags` (unlike `--add-flag`) skips `@Q` quoting — `flags="${params[...]}"`, then the final `echo exec ... "${flagsBefore-}" '"$@"'` writes it verbatim into the wrapper.
- Result: `exec -a "$0" .../.ChatGPT-wrapped ${WAYLAND_DISPLAY:+--ozone-platform=wayland} "$@"`, unquoted, so it expands at launch and word-splits away to nothing when unset. `${x:+y}` is also safe under the wrapper's `#! bash -e` (and would be under `set -u`).

Dropping `ELECTRON_OZONE_PLATFORM_HINT` loses nothing on X11 — `auto` resolves to x11 exactly when `WAYLAND_DISPLAY` is unset, which is the same condition. It does drop the env-var escape hatch (`ELECTRON_OZONE_PLATFORM_HINT=x11` to force XWayland). Still reachable via `--ozone-platform=x11` in `~/.config/chatgpt-flags.conf`: `codex-launcher` puts conf flags after the wrapper's, and Chromium's switch map is last-one-wins. Might be worth a half-sentence in the comment above, since the flags file is now the only way.

Nit, on the PR's own thesis: `--prefix PATH` now puts a store `coreutils` ahead of the user's PATH for every child, including the `codex` agent shelling out in the user's project. `--suffix` would get `cp`/`chmod` resolved just as reliably without reordering anything for the agent.

---

pkgs/chatgpt-desktop/patch-asar.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python3
+"""Apply byte-length-preserving source patches inside app.asar.
+
+Vendored from numtide/llm-agents.nix (MIT), which hit the same two problems.
```
cprussin: The file is byte-identical to upstream apart from this line, and the mechanics check out: both patterns are anchored on captured identifiers rather than minified names, both `sys.exit` on a match count other than 1 or on an over-long replacement, and space padding is safe in both contexts (`false /* nix:skip report */` is exactly the 27 bytes of `isLinux() && process.report`). `cp -r src/. dst` also does the right thing whether or not `dst` exists, and preserves symlinks as symlinks like the `verbatimSymlinks: !0` it replaces — I checked against GNU coreutils 9.4. The `chmod -R u+w` is genuinely required, since `cp` gives the created top-level directory the source's `r-xr-xr-x`.

Two things:

**Attribution.** MIT wants the copyright notice and permission notice carried into copies; "Vendored from numtide/llm-agents.nix (MIT)" is a pointer, not either of those. Worth adding the upstream copyright line, plus a permalink at the exact revision rather than the repo — it's the only way to diff against upstream later when a version bump moves the minified code and this file needs re-syncing.

**Behavior narrowing worth a comment.** `e + '/.'` makes the helper directory-only, where the `fs.cp` and darwin `ditto` branches it sits beside both handle a plain file. Upstream has the same shape, so presumably every caller passes a directory today — but `cp: '.../x/.' is not a directory` is a confusing failure if that stops being true, and the two `sys.exit` guards above won't catch it (it's a runtime failure, not a build one).

---

cprussin (commented):
Both fixes are the right shape — `runtimeDependencies` instead of a wrapper-level `LD_LIBRARY_PATH` is exactly how nixpkgs handles prebuilt Electron, and the asar patch fails the build loudly if the code it matches moves.

One thing to fix, one follow-up, some notes inline. (Submitting as a comment rather than request-changes only because GitHub won't let me request changes on your own PR — treat the first item as blocking.)

**Blocking:** the `runtimeDependencies` set is missing `libpulseaudio` and `libudev` (via `lib.getLib systemd`). Both were covered by the old `LD_LIBRARY_PATH`, both are `dlopen`ed by Chromium so they'll never show up in a `DT_NEEDED` for autoPatchelf to resolve, and both fail silently — no audio (voice mode), no device enumeration. Details and the `lib.getLib` trap in the inline comment.

**Verified, no change needed:** the `--add-flags` shell expansion does survive `wrapProgram` — `--add-flags` deliberately skips the `@Q` quoting that `--add-flag` applies, so the wrapper gets an unquoted `${WAYLAND_DISPLAY:+...}` that expands at launch. Nothing is lost on X11 either. Walked through inline.

**Follow-up (not this PR):** `pkgs/claude-desktop/derivation.nix:209-210` still has the `--prefix LD_LIBRARY_PATH` + `ELECTRON_OZONE_PLATFORM_HINT auto` pair this PR is removing. That app also spawns a browser for sign-in and runs the `claude` CLI as a child, so the same leak should produce the same two symptoms. If the ChatGPT fix holds up on hardware, it should port over.

`deadnix`/`statix`: clean in both directions — `coreutils`, `pipewire` and `python3` are all declared and all used, and nothing became unused (`runtimeLibs` still feeds `buildInputs`, `xdg-utils` moved into `runtimeBins`).

Caveat on all of the above: no nix in this environment and `persistent.oaistatic.com` is egress-blocked, so nothing here is empirically verified against the actual binaries — it's read against nixpkgs sources at the pinned rev, numtide's `unwrapped.nix`, and the AUR PKGBUILD.

---

pkgs/chatgpt-desktop/derivation.nix
```diff
@@ -199,16 +237,24 @@ in
...
     '';
 
+    # `wrapProgramShell`, not `wrapProgram`: wrapGAppsHook3 propagates
+    # makeBinaryWrapper, which redefines `wrapProgram` to emit a C wrapper that
+    # passes --add-flags through verbatim -- the `${WAYLAND_DISPLAY:+...}` above
+    # would reach Electron as a literal argument instead of expanding.  The
+    # shell flavour stays available because makeWrapper is in nativeBuildInputs.
     postFixup = ''
-      wrapProgram "$out/lib/chatgpt/ChatGPT" "''${gappsWrapperArgs[@]}"
+      wrapProgramShell "$out/lib/chatgpt/ChatGPT" "''${gappsWrapperArgs[@]}"
```
cprussin: **`--inherit-argv0` does not survive the shell flavour — argv[0] changes from `chatgpt` to `ChatGPT`.**

I verified everything the comment claims, and the premise holds. But the swap has one behavioural consequence the comment doesn't mention, and it lands squarely on the thing you're asking to be verified with `swaymsg`.

`wrapProgramShell` calls `makeShellWrapper "$hidden" "$prog" --inherit-argv0 …` (`make-wrapper.sh:234`), and `--inherit-argv0` in the shell flavour is implemented as `argv0='$0'` (`make-wrapper.sh:188-190`), emitted as `exec -a "$0" …` (`:199-200`). In a shebang script `$0` is **the path the kernel used to exec the script**, not the caller's argv[0] — the shebang mechanism discards the caller's argv[0] entirely:

```
$ printf '#!/bin/bash\necho "dollar0=$0"\n' > /tmp/t/s.sh; chmod +x /tmp/t/s.sh
$ bash -c 'exec -a FAKENAME /tmp/t/s.sh'
dollar0=/tmp/t/s.sh
```

The C wrapper genuinely inherits argv[0]. So:

- **before:** `codex-launcher` execs its sibling with its own argv[0] (`$out/bin/chatgpt` from the `.desktop` `Exec=`), the binary wrapper passes it through → Chromium sees program name `chatgpt`.
- **after:** bash's `$0` is `$out/lib/chatgpt/ChatGPT` → Chromium sees program name `ChatGPT`.

Chromium derives `WM_CLASS` / the Wayland `app_id` from the program name when `--class` isn't given, so this is a plausible route to `app_id = "ChatGPT"`, which won't match `chatgpt.desktop` and loses the icon association — i.e. the check in your PR body could come back "an app_id, but the wrong one".

Cheap fix: append `--argv0 chatgpt` to `gappsWrapperArgs`. `makeShellWrapper` processes args in order and "whichever comes last of `--argv0` and `--inherit-argv0` wins" (`make-wrapper.sh:189`), so a later `--argv0` overrides the `--inherit-argv0` that `wrapProgramShell` injects. `--add-flags "--class=chatgpt"` would also pin it explicitly.

---

pkgs/chatgpt-desktop/derivation.nix
```diff
@@ -130,11 +140,32 @@ in
...
+      libpulseaudio
+      libsecret
+      pipewire
+      vulkan-loader
+      wayland
+      # systemd has no `lib` output at this nixpkgs rev, so this resolves to
+      # the same path as bare `systemd` and is only here to stay correct if one
+      # is ever added.  libudev.so.1 does live in its $out/lib, which is where
+      # the hook looks: it appends /lib to whatever path it is given.
+      (lib.getLib systemd)
```
cprussin, line 166: Verified accurate, all three claims, against nixpkgs `d407951447dcd00442e97087bf374aad70c04cea` (the rev `flake.lock` pins for the `nixpkgs` node).

1. "no `lib` output at this nixpkgs rev" — `pkgs/os-specific/linux/systemd/default.nix:269`: `outputs = [ "out" "dev" ] ++ lib.optional (!buildLibsOnly) "man"`. `lib.getLib` is `getOutput "lib"`, which falls back to `.out` when the attr is absent, so this is the same store path as bare `systemd`.
2. "libudev.so.1 does live in its `$out/lib`" — there is no `moveToOutput` for it anywhere in that file, and no `libdir` override in `mesonFlags` (the only libdir touched is `pkgconfiglibdir` → `$dev`, line 424), so meson's nixpkgs default `--libdir=lib` puts it at `$out/lib`.
3. "the hook … appends /lib to whatever path it is given" — `pkgs/by-name/au/autoPatchelfHook/auto-patchelf.sh:80`: `--runtime-dependencies "${runtimeDependenciesArray[@]/%//lib}"`.

Nothing left to change here.

---

pkgs/chatgpt-desktop/derivation.nix
```diff
@@ -187,28 +218,45 @@ in
...
-      # own path, so wrapping the sibling in place below is what gets the
-      # environment onto it.
+      # beside it: a two-line script that resolves its sibling from its own
+      # path and execs it.  The wrapper goes on that sibling because ChatGPT is
+      # the process that needs the flags and the environment, and wrapping it
+      # also covers anything that runs it directly rather than via the
+      # launcher.
       test -e "$out/lib/chatgpt/codex-launcher" \
         || (echo "no codex-launcher in this .deb -- wrap ChatGPT directly" >&2; exit 1)
       ln -s ../lib/chatgpt/codex-launcher "$out/bin/chatgpt"
```
cprussin, line 237: The rewrite is sound: it no longer asserts anything about `readlink -f` undoing the wrapper, and what it does assert is either verifiable or a plain statement of intent ("ChatGPT is the process that needs the flags and the environment", "also covers anything that runs it directly").

The one claim I could not check first-hand is "a two-line script" — the `.deb` host is not reachable from this sandbox. But the structure is corroborated by the package this file's asar patch is vendored from, `numtide/llm-agents.nix` `packages/chatgpt/unwrapped.nix`, which does exactly the same two things: symlinks `$out/bin/chatgpt` → `../lib/chatgpt/codex-launcher` (line 153) and wraps `$out/lib/chatgpt/ChatGPT` (line 158). Two independent packagers landing on the same launcher/sibling split is good evidence the description is right.

---

pkgs/chatgpt-desktop/derivation.nix
```diff
@@ -130,11 +140,32 @@ in
...
       dpkg
       makeWrapper
+      python3
       wrapGAppsHook3
     ];
 
     buildInputs = runtimeLibs;
 
+    # Electron dlopens these rather than linking them, so autoPatchelf has to
+    # be told: they go on each ELF object's RPATH.  What they must not do is
```
cprussin: Optional, non-blocking wording point, since comment accuracy is what the last two rounds have been about.

"each ELF object's RPATH" is slightly wider than what the hook does — `runtimeDependencies` land only on dynamically-linked **executables**, not on `.so` files. `pkgs/by-name/au/auto-patchelf/source/auto-patchelf.py:351-357` at `d407951`:

```python
rpath = []
if file_is_dynamic_executable:
    ...
    rpath += runtime_deps
```

Doesn't change anything functionally here — the dlopens in question all happen from the `ChatGPT` executable — and the phrasing is inherited from the upstream `llm-agents.nix` comment, so ignore this if you'd rather stay close to that. "each executable's RPATH" would just be exactly true.

---

pkgs/chatgpt-desktop/derivation.nix
```diff
@@ -130,11 +140,32 @@ in
...
+    # arrive as an LD_LIBRARY_PATH on the app's environment -- see the wrapper
+    # below.
+    runtimeDependencies = [
```
cprussin, line 153: Re: `systemd` appearing both here and in `runtimeLibs` (→ `buildInputs`, line 125) — your reasoning holds, and it is not even a special case.

The two lists feed different arguments and neither substitutes for the other. `buildInputs` becomes `--libs` via the `gatherLibraries` env hook (`auto-patchelf.sh:6-12, 78-79`), and only those paths reach `populate_cache`, which builds the soname index `find_dependency` resolves `DT_NEEDED` against (`auto-patchelf.py:460-462`). `--runtime-dependencies` is passed straight through to the per-file RPATH and is never added to the search path. So dropping `systemd` from `buildInputs` really would make any `DT_NEEDED libudev.so.1` / `libsystemd.so.0` unresolvable — a build failure, exactly as you said.

Also worth noting it's the file's uniform pattern rather than an oddity: `libGL`, `libgbm`, `libnotify`, `libpulseaudio`, `libsecret`, `vulkan-loader` and `wayland` are all in both lists too. `pipewire` is the only entry in `runtimeDependencies` alone (upstream `llm-agents.nix` has it in `buildInputs` as well), which is fine — if something ever gained a direct link to it the build would fail loudly rather than silently.

Leave it as is.

---

cprussin (commented):
Reviewed the `wrapProgram` → `wrapProgramShell` swap against nixpkgs at the pinned rev `d407951447dcd00442e97087bf374aad70c04cea`, plus the rest of the diff. **The change is correct and it fixes a real bug** — the round-2 note on the previous commit that said "the `--add-flags` shell expansion does survive `wrapProgram`, no change needed" was wrong, and you were right to reverse it. One real regression came in with it, one comment is factually incorrect, and there's a simpler alternative worth ruling out first.

(GitHub's GraphQL API has been rate-limited for this account throughout this review, so I could only land one of four inline comments. The remaining three are inlined below with file:line.)

---

## Premises: all verified

- **`wrapGAppsHook3` does propagate `makeBinaryWrapper`.** `all-packages.nix:892-893` at the pin is `wrapGAppsNoGuiHook = callPackage ../build-support/setup-hooks/wrap-gapps-hook { makeWrapper = makeBinaryWrapper; }`, `wrapGAppsHook3` is that with `isGraphical = true` (`:883-885`), and `wrap-gapps-hook/default.nix:18-22` puts that `makeWrapper` in `propagatedBuildInputs`.
- **The binary flavour did win.** Setup hooks activate in `nativeBuildInputs` order with propagated deps spliced in during `findInputs` (setup.sh `_activatePkgs`), so the order is `autoPatchelfHook → dpkg → makeWrapper → python3 → wrapGAppsHook3 → makeBinaryWrapper`. `make-binary-wrapper.sh:61` (`wrapProgram() { wrapProgramBinary "$@"; }`) is sourced last and wins.
- **The C wrapper really would have passed the parameter expansion through verbatim.** `make-binary-wrapper.sh:168-173` does `read -ra flags <<< "$arg"` and `addFlags` (`:224-247`) bakes each word in as an escaped C string literal — `argv_tmp[1] = "${WAYLAND_DISPLAY:+--ozone-platform=wayland}";` — with no shell anywhere in the path. Electron would have seen that literal as argv[1]. This was a genuine defect.
- **`wrapProgramShell` is in scope and nothing clobbers it.** `make-binary-wrapper.sh` defines only `makeWrapper`, `makeBinaryWrapper`, `wrapProgram`, `wrapProgramBinary` and helpers; it neither defines nor unsets `wrapProgramShell`/`makeShellWrapper`, so `make-wrapper.sh:222-235` survives intact. Naming the flavour explicitly is also strictly more robust than depending on that sourcing order.
- **No argument-syntax hazard in either direction.** `gappsWrapperArgs` only ever gets `--set` and `--prefix ENV : VAL` (`wrap-gapps-hook.sh:6,14,18,23,27,32`), both handled identically. More generally the binary wrapper's accepted set (`--set`, `--set-default`, `--unset`, `--prefix`, `--suffix`, `--chdir`, `--add-flag(s)`, `--append-flag(s)`, `--argv0`, `--inherit-argv0`, `--resolve-argv0`) is a strict **subset** of the shell wrapper's, which additionally has `--run`, `--prefix-each`, `--suffix-each` and `--{prefix,suffix}-contents`. Binary → shell cannot hit an unsupported arg.
- **Expansion lands where intended.** `make-wrapper.sh:199-200` writes `flagsBefore` unquoted into the script, so the emitted line is `exec -a "$0" ".../.ChatGPT-wrapped" ${WAYLAND_DISPLAY:+--ozone-platform=wayland} "$@"`. Unset → unquoted expansion produces zero words, so X11 sessions are untouched. Correct.
- **Phase ordering is fine.** `runHook postFixup` evaluates the derivation's `postFixup` string *before* `postFixupHooks` (setup.sh:198-210 via `_callImplicitHook`), and `auto-patchelf.sh:104` registers there — so auto-patchelf sees the ELF already renamed to `.ChatGPT-wrapped`. It handles that (pathlib's `rglob('*')` includes dotfiles, unlike the `glob` module), and the CI log for this commit confirms it: `setting interpreter of .../lib/chatgpt/.ChatGPT-wrapped` followed by full NEEDED resolution and the `runtimeDependencies` RPATH.
- **Costs are nil.** bash `exec`s, so no lingering process and no signal-forwarding problem; ~1ms of startup; no setuid anywhere here.

## Finding 1 (real regression) — `pkgs/chatgpt-desktop/derivation.nix:257`

Left as an inline comment. Summary: `wrapProgramShell` injects `--inherit-argv0` (`make-wrapper.sh:234`), which the shell flavour implements as `argv0='$0'` → `exec -a "$0"` (`:188-190`, `:199`). In a shebang script `$0` is the path the *kernel* used, not the caller's argv[0] — the shebang mechanism discards the caller's argv[0] outright:

```
$ bash -c 'exec -a FAKENAME /tmp/t/s.sh'    # s.sh echoes "$0"
dollar0=/tmp/t/s.sh
```

The C wrapper genuinely inherits it (`make-binary-wrapper.sh:203`: with `inherit_argv0` set it simply never assigns `argv[0]`). So argv[0] reaching Electron changes from whatever `codex-launcher` passed — presumably `$out/bin/chatgpt`, basename `chatgpt` — to `$out/lib/chatgpt/ChatGPT`, basename `ChatGPT`. Chromium derives `WM_CLASS`/the Wayland `app_id` from the program name absent `--class`, so your `swaymsg -t get_tree | grep -i chatgpt` check could come back with an `app_id` that no longer matches `chatgpt.desktop` and loses the icon association. Fix: append `--argv0 chatgpt` to `gappsWrapperArgs` (a later `--argv0` beats the injected `--inherit-argv0`, `make-wrapper.sh:189`), or pin it with `--add-flags "--class=chatgpt"`.

## Finding 2 (worth ruling out first) — `derivation.nix:240-249`

`config/modules/ui/launcher/default.nix:16,17,90` already pass `--ozone-platform-hint=auto` to element-desktop, Telegram and Slack. That's a Chromium *flag*, distinct from the `ELECTRON_OZONE_PLATFORM_HINT` env var this fork reportedly ignores; it does exactly the auto-Wayland-else-X11 selection being hand-rolled here, and being a literal flag it survives the binary wrapper fine — no `wrapProgramShell`, no `makeWrapper` in `nativeBuildInputs`, no argv0 change. The PR body says items 2 and 3 are unverified on hardware, so "this fork ignores the hint" is a hypothesis and whether it also ignores the *flag* is untested. Worth trying before taking on the wrapper swap.

For contrast, numtide — whose `patch-asar.py` you vendored — does neither: they keep `wrapProgram` (binary) on the inner `ChatGPT` and put the shell expansion in a separate *outer* `makeShellWrapper` around `bin/chatgpt` (`packages/chatgpt/package.nix:22-24`), gated on `NIXOS_OZONE_WL` as well as `WAYLAND_DISPLAY`.

That outer placement is **not** a drop-in for what you have, though: `codex-launcher` prepends the conf-file flags, so flags handed to `bin/chatgpt` land *after* them and `chatgpt-flags.conf` would lose (base::CommandLine is last-wins on duplicate switches). Your inner placement puts `--ozone-platform=wayland` before the conf flags, so `--ozone-platform=x11` there wins — exactly what lines 242-243 claim, and it follows from `make-wrapper.sh:199-200` emitting `--add-flags` content as `flagsBefore`, ahead of `"$@"`. The comment is right and the placement is the better of the two; it's only the wrapper *flavour* in question.

Separately: `WAYLAND_DISPLAY` alone forces native Wayland even for someone who has deliberately opted out of Ozone-Wayland globally. Your call for a personal config — the repo has no other `NIXOS_OZONE_WL` usage to be consistent with either way.

## Finding 3 (comment is wrong) — `derivation.nix:162-165`

> `lib.getLib` matters here and nowhere else in this list: the hook appends /lib to whatever path it's given, and systemd's default output keeps units and udev rules there rather than libudev.so.1.

systemd at this pin declares `outputs = [ "out" "dev" ] ++ optional (!buildLibsOnly) "man"` (`pkgs/os-specific/linux/systemd/default.nix:269-273`) — **there is no `lib` output**, so `lib.getLib systemd` is exactly `systemd`, i.e. the default output the comment says it is avoiding. The CI log confirms it: the RPATH entry is `/nix/store/0xiinbh7pbj479i5v28mqnjrvhvbdibf-systemd-260.2/lib`, with no `-lib` output suffix. It works only because systemd's own `$out/lib` holds `libudev.so.1` right alongside the units.

Two things worth doing:
1. Fix or drop the comment — as written it documents behaviour that doesn't exist.
2. Consider `systemdLibs`/`systemdMinimal` instead. Auto-patchelf already resolved the `DT_NEEDED` `libudev.so.1` to `systemd-minimal-libs` in the same build, so this `runtimeDependencies` entry is the only thing dragging full systemd into the closure — and now that the `LD_LIBRARY_PATH` that used to reference it is gone, switching would actually drop it.

## Finding 4 (minor, unremarked behaviour change) — `derivation.nix:246`

`--prefix PATH` became `--suffix PATH`. I think this is deliberate and correct — suffixing keeps a pinned `coreutils` out of front position for every child the app spawns (`codex`, its shell tool calls, `xdg-open` and the browser), which is exactly this PR's theme. But it's undocumented, it diverges from numtide's `--prefix` (`unwrapped.nix:158-166`), and it diverges from its sibling in this repo (`pkgs/claude-desktop/derivation.nix:208`). It also sits slightly oddly against the comment at `:79-80` calling coreutils "load-bearing", when in practice the user's PATH supplies `cp`/`chmod` and the store copy is only a fallback. Worth a clause saying that's the intent.

## Finding 5 (minor, inherited from upstream) — `patch-asar.py:68`

`cp -r e/. t` and node's `fs.cp(e, t, {recursive: true})` differ in one way: `fs.cp` creates the whole destination chain, `cp` creates only the final component.

```
$ cp -r src/. dst          # dst absent  → OK, dst created
$ cp -r src/. a/b/c        # a/ absent   → cp: cannot create directory 'a/b/c'
```

So a first launch against a clean `~/.codex` fails if the app relied on `fs.cp` to `mkdir -p` an intermediate level. It's vendored verbatim from a package that presumably works, so likely a non-issue — flagging it because it's the one behavioural gap between the two calls and first launch is exactly when it would bite. Everything else about that patch checks out: `cp -r` copies symlinks as symlinks (matching the dropped `verbatimSymlinks: true`) and applies the source mode, which is precisely why the follow-up `chmod -R u+w` is needed.

## Everything else: no issues

- `patch-asar.py` is a faithful vendor — `diff` against `numtide/llm-agents.nix` `packages/chatgpt/patch-asar.py` shows only the added provenance/MIT header. The byte-length-preserving approach is sound for asar (the header records offsets), the `!= 1` match count and the length guard both hard-fail the build, and `sys.exit(str)` gives a non-zero exit with the message on stderr.
- `runtimeDependencies` (`:153-166`) looks complete against what the removed `LD_LIBRARY_PATH` covered. I checked the CI log's `DT_NEEDED` resolution for `.ChatGPT-wrapped`: everything else in `runtimeLibs` (glib, nspr/nss, at-spi2, dbus, cups, expat, libxcb, libxkbcommon, alsa, X libs, cairo, pango, gdk-pixbuf, gtk3, libudev) is a real NEEDED entry and gets patched without help. The nine listed here are the dlopen-only ones. `libpulseaudio` and `libudev`, flagged as blocking in the previous round, are both present now.
- `autoPatchelfIgnoreMissingDeps` (`:173-184`) matches the build exactly — the 11 warnings emitted are the six Qt shims and five `libc.musl-x86_64.so.1`. `libc++_shared.so`, `libc.musl-aarch64.so.1` and `liblog.so` don't fire on x86_64 (those prebuilds are skipped by architecture) but are needed for the aarch64 platform; unused entries are harmless. Enumerating rather than blanket-ignoring is the right call.
- `installPhase`: `python3 ${./patch-asar.py}` is correctly placed after the copy and before `fixupPhase`, and `tar --no-same-permissions` at `:193` leaves `app.asar` writable, so the in-place rewrite works without an extra `chmod`. The `codex-launcher` existence guard at `:233-234` still holds.
- No launcher-module or lyra-dataset changes in this PR — it's two files.

Caveat: no nix in this environment, so the derivation itself wasn't evaluated. Findings 1-4 are read against nixpkgs sources at the pin and cross-checked against the `build pkgs.chatgpt-desktop` job log for `56d3d4b` (all 16 checks green); finding 5 and the `cp`/`$0` behaviours were reproduced locally.

---

cprussin (commented):
Addendum — the branch was force-pushed from `56d3d4b` to `653e764` while I was blocked on GitHub's GraphQL rate limit, so two corrections to the review I just posted:

- **Finding 3 is already fixed.** The `lib.getLib systemd` comment at `derivation.nix:162-165` now reads "systemd has no `lib` output at this nixpkgs rev, so this resolves to the same path as bare `systemd`…", which is exactly right. Ignore that half of the finding. The other half stands and is unaddressed: auto-patchelf resolves the `DT_NEEDED` `libudev.so.1` to `systemd-minimal-libs`, so this `runtimeDependencies` entry is the sole thing pulling full `systemd` into the closure now that the `LD_LIBRARY_PATH` is gone — `systemdLibs`/`systemdMinimal` would drop it.
- **Finding 2's precedence argument now rests on a different premise.** The `installPhase` comment no longer claims `codex-launcher` reads `chatgpt-flags.conf` and prepends those flags; it now just says the launcher resolves its sibling and execs it, and the `preFixup` comment says "Being flagsBefore, a `--ozone-platform=` passed on the command line still wins". That's true for anything appearing after `"$@"`-position args, so the conclusion is unchanged — but if `codex-launcher` really does still inject `chatgpt-flags.conf` contents, it'd be worth keeping that in the comment, since it's the only reason the conf file can override the wrapper's flag.

Findings 1, 2, 4 and 5 all apply unchanged to `653e764` — `wrapProgramShell` and the wrapper args are byte-identical between the two commits; only comments moved.

---

cprussin:
Responses to the review on `56d3d4b`. Findings 3 and 4 are addressed in `c3bb33d`; findings 1, 2 and 5 I'm not acting on, with reasons below.

## Finding 1 (`--inherit-argv0` regression) — not a regression

The mechanism is right but the baseline is wrong. The claim is that argv[0] shifts from `chatgpt` to `ChatGPT`. It was never `chatgpt`: `codex-launcher` is

```sh
exec "$(dirname "$(readlink -f "$0")")/ChatGPT" "$@"
```

`exec "$prog"` with no `-a` sets argv[0] to `$prog` — the resolved sibling path — so the binary wrapper's `--inherit-argv0` was inheriting `…/lib/chatgpt/ChatGPT`, not anything derived from `bin/chatgpt`.

Reproduced with a real ELF (a shell script as the stand-in doesn't work here — the shebang re-exec discards the spoofed argv0, which is what makes this easy to get wrong):

```
== SHELL wrapper (this PR), via bin/chatgpt, WAYLAND_DISPLAY set:
  argv0=/tmp/…/out/lib/chatgpt/ChatGPT
  arg1=--ozone-platform=wayland
== SHELL wrapper, no WAYLAND_DISPLAY:
  argv0=/tmp/…/out/lib/chatgpt/ChatGPT
== BINARY wrapper (--inherit-argv0), same launcher chain:
  argv0=/tmp/…/out/lib/chatgpt/ChatGPT-binflavour
```

Basename is `ChatGPT` either way, so `WM_CLASS`/`app_id` is unchanged by this PR. `--argv0 chatgpt` would be a *new* behaviour change, unverified against whatever `StartupWMClass` the shipped `.desktop` carries, so it's not going in here.

That run is also the first empirical confirmation of the fix itself: the flag appears under `WAYLAND_DISPLAY` and produces zero words without it.

## Finding 2 (`--ozone-platform-hint=auto`) — noted, not switching blind

Fair point on repo consistency, and it would avoid the wrapper-flavour question entirely. But it trades a mechanism I've now verified end-to-end for one that's untested against this particular Electron fork — and the premise for the whole change is that the fork ignores the hint in its env-var form. Whether it honours the switch form is a separate unknown. I'd rather land the verified version and revisit if hardware testing shows `--ozone-platform=wayland` isn't taking effect.

## Finding 3 (`lib.getLib systemd` comment) — fixed

Fixed in `653e764` (the force-push you saw mid-review). The comment now says systemd has no `lib` output at this pin, that this therefore resolves to the same path as bare `systemd`, and that `libudev.so.1` does live in its `$out/lib`.

Not taking the `systemdLibs` half. It's a real closure win, but it's a packaging change unrelated to this PR's subject and belongs in its own change where it can be evaluated on its own.

## Finding 4 (`--prefix` → `--suffix` undocumented) — fixed

`c3bb33d` adds the rationale in the code, not just the commit message, including the divergence from `pkgs/claude-desktop/derivation.nix`.

## Finding 5 (`cp -r` vs `fs.cp` intermediate dirs) — real, but unfixable as stated

Correct that `cp -r src/. a/b/c` fails where `fs.cp` would `mkdir -p`. It can't be fixed in `patch-asar.py` as written: the patch is byte-length-preserving because asar's header records absolute offsets, and the replacement is already the same length as the original. Adding a `mkdir -p` would lengthen it and corrupt the archive.

Leaving it, and noting why: the app creates its own staging directories (the `EACCES` messages in the failure this patch fixes are on `open`/`rmdir` of paths that already exist), and this is vendored verbatim from a package in production use.

---

cprussin:
On the addendum's second point — `codex-launcher` does **not** inject `chatgpt-flags.conf`. That claim was in an earlier version of the comment and was wrong; removing it is why the comment changed. Read off the deployed store path:

```
$ cat /nix/store/…-chatgpt-desktop-26.820.71523/lib/chatgpt/codex-launcher
#!/nix/store/…-bash-5.3p9/bin/sh
exec "$(dirname "$(readlink -f "$0")")/ChatGPT" "$@"
```

That's the whole file. So there's nothing to restore to the comment, and the flag-precedence claim rests only on `flagsBefore` ordering plus Chromium's last-wins switch parsing — both of which this review already verified independently.

Both halves of finding 3 are answered: the comment is fixed in `653e764`, and the `systemdLibs` closure win is real but out of scope here, as noted above.

---

PR #21, opened 2026-08-28
[launcher] Stop the ChatGPT app answering for every web link

cprussin:
Fixes the ChatGPT sign-in failure, which turned out not to be a packaging problem with the app at all.

## What was happening

Clicking **Sign in** did nothing — a flash of "Continue to sign in on your browser", then back to the sign-in screen, and "A ChatGPT login is already in progress" on a retry.

`strace -f -e trace=execve` on the app while clicking:

```
execve(".../coreutils/bin/env", [..., ".../chatgpt-desktop-26.820.71523/bin/chatgpt",
                                 "https://chatgpt.com/codex/deskto"...])
execve(".../chatgpt-desktop-26.820.71523/bin/chatgpt", [..., "https://chatgpt.com/codex/deskto"...])
```

The app resolves the login URL through the xdg handler for `https` and execs that `.desktop`'s `Exec` line — which is itself. The second instance finds the first through Chromium's `ProcessSingleton`, prints `Opening in existing browser session.`, and exits. No browser is launched, and the first login stays pending forever.

It answered for `https` because the `.deb` declares:

```
MimeType=x-scheme-handler/codex;x-scheme-handler/http;x-scheme-handler/https;text/csv;...
```

and this machine has no default browser registered:

```
$ xdg-settings get default-web-browser
                     # empty
$ grep -rn x-scheme-handler/http ~/.config/mimeapps.list
                     # nothing
```

With no default set, the only claimant wins. So this was never only a login bug — *any* app opening a web link here reached the ChatGPT app.

## What this changes

**The app stops claiming the two web schemes.** It keeps `x-scheme-handler/codex`, which is genuinely its own, and the file-type associations. `--replace-fail`, so a future `.deb` that stops shipping that exact list fails the build rather than silently going back to answering for the web.

**A default gets registered**, because once ChatGPT stops claiming `https` nothing else here does. `$BROWSER` is already the launcher's `browse` script, but `$BROWSER` only reaches things that read it, which desktop apps mostly don't. The same script now gets a `.desktop` and is set as the default for `http`, `https` and `text/html`, so both routes land in one place and `browse`'s own `--browser` default stays the single choice of browser.

## Ruled out along the way

- **`LD_LIBRARY_PATH` leaking into children** — real, and fixed in #19, but not this. Running the bundled `resources/codex login` by hand opens a browser with the app's `LD_LIBRARY_PATH` set and without it alike.
- **`xdg-open`** — never invoked. A logging shim ahead of the app on `PATH` was never called.
- **The D-Bus portal** — `dbus-monitor` during a click shows no `OpenURI` call. The only portal traffic is `Settings.ReadAll` for theme detection, which fails `UnknownMethod` on every GTK app here and is unrelated.

## Verification

CI covers evaluation and the package build. The behavioural check needs hardware: after deploying, `xdg-settings get default-web-browser` should report `browse.desktop`, and clicking **Sign in** should open a browser.

## Noticed, not fixed here

`xdg.portal.extraPortals` at `config/modules/ui/sway/default.nix:205` installs nothing. home-manager's portal module adds backends via `home.packages` at normal priority, and this repo's `lib.mkForce` idiom on `home.packages` discards them — which is why only `xdg-desktop-portal-wlr` (hand-added to a `mkForce` list) is on the bus and `gtk` is not. Unrelated to this bug, but it means `FileChooser` and `OpenURI` have no implementation on this machine. Worth its own change.

https://claude.ai/code/session_013VHH9pj1eDhtmmfNwE4ZnY

---
_Generated by [Claude Code](https://claude.ai/code/session_013VHH9pj1eDhtmmfNwE4ZnY)_

### DIFF
--- a/pkgs/chatgpt-desktop/derivation.nix
+++ b/pkgs/chatgpt-desktop/derivation.nix
@@ -212,10 +212,17 @@ in
         fi
       done
 
+      # The app claims x-scheme-handler/http and https because it has an
+      # in-app browser.  On a machine with no default browser set that makes it
+      # the handler for every web link, and the app resolves the login URL that
+      # way -- so clicking "sign in" re-execs the app, which hands off to the
+      # running instance ("Opening in existing browser session.") and exits,
+      # and no browser ever opens.  It should not be a general web browser.
       for desktop in "$out"/share/applications/*.desktop
       do
         substituteInPlace "$desktop" \
-          --replace-fail "Exec=chatgpt" "Exec=$out/bin/chatgpt"
+          --replace-fail "Exec=chatgpt" "Exec=$out/bin/chatgpt" \
+          --replace-fail "x-scheme-handler/http;x-scheme-handler/https;" ""
       done
 
       # Two things the app does that only break on NixOS: it materializes its
--- a/pkgs/chatgpt-desktop/derivation.nix
+++ b/pkgs/chatgpt-desktop/derivation.nix
@@ -218,6 +218,13 @@ in
       # way -- so clicking "sign in" re-execs the app, which hands off to the
       # running instance ("Opening in existing browser session.") and exits,
       # and no browser ever opens.  It should not be a general web browser.
+      #
+      # This package is in no profile, so the entry here never reaches the MIME
+      # database directly: the app self-installs a copy into
+      # ~/.local/share/applications, which is what was answering.  Patching the
+      # source of that copy governs future ones only -- a copy already there
+      # keeps its claim until the app rewrites it -- so the registered default
+      # in the launcher module is what actually settles the handler.
       for desktop in "$out"/share/applications/*.desktop
       do
         substituteInPlace "$desktop" \
--- a/pkgs/launcher/home-manager-module.nix
+++ b/pkgs/launcher/home-manager-module.nix
@@ -36,19 +36,35 @@ in {
     # app claims it for its in-app browser, so it was answering for every web
     # link on the machine.  Point both routes at the same script instead.
     xdg = {
-      desktopEntries.browse = {
-        name = "Browse";
+      # `xdg.dataFile` rather than `xdg.desktopEntries`, which would be the
+      # obvious option: it installs through `home.packages`, at the normal
+      # priority that every `lib.mkForce [...]` on that list in this repo
+      # discards -- including the one in the module that imports this.  The
+      # entry would silently not exist, and mimeApps below would then point
+      # every web link at nothing at all.  This lands it in
+      # ~/.local/share/applications, which the spec searches ahead of
+      # XDG_DATA_DIRS, so it also doesn't lean on `environment.pathsToLink`.
+      #
+      # %u, not %U: `browse` keeps one target and drops the rest, so claiming
+      # to accept a list would silently lose all but the last URL.
+      dataFile."applications/browse.desktop".source = "${pkgs.makeDesktopItem {
+        name = "browse";
+        desktopName = "Browse";
         genericName = "Web Browser";
-        exec = "${pkgs.launcher}/bin/browse %U";
+        exec = "${pkgs.launcher}/bin/browse %u";
         terminal = false;
         categories = ["Network" "WebBrowser"];
-        mimeType = [
+        mimeTypes = [
           "text/html"
           "x-scheme-handler/http"
           "x-scheme-handler/https"
         ];
-      };
+      }}/share/applications/browse.desktop";
 
+      # This makes ~/.config/mimeapps.list a read-only store symlink, so
+      # `xdg-mime default`, `xdg-settings set default-web-browser` and a
+      # browser's own "make me the default" prompt all stop working.  Defaults
+      # get set here and redeployed, not set at runtime.
       mimeApps = {
         enable = true;
         defaultApplications = {
--- a/pkgs/launcher/home-manager-module.nix
+++ b/pkgs/launcher/home-manager-module.nix
@@ -47,6 +47,9 @@ in {
       #
       # %u, not %U: `browse` keeps one target and drops the rest, so claiming
       # to accept a list would silently lose all but the last URL.
+      configFile."mimeapps.list".force = true;
+      dataFile."applications/mimeapps.list".force = true;
+
       dataFile."applications/browse.desktop".source = "${pkgs.makeDesktopItem {
         name = "browse";
         desktopName = "Browse";
@@ -61,10 +64,16 @@ in {
         ];
       }}/share/applications/browse.desktop";
 
-      # This makes ~/.config/mimeapps.list a read-only store symlink, so
-      # `xdg-mime default`, `xdg-settings set default-web-browser` and a
-      # browser's own "make me the default" prompt all stop working.  Defaults
-      # get set here and redeployed, not set at runtime.
+      # This makes ~/.config/mimeapps.list a read-only store symlink (and the
+      # deprecated copy under ~/.local/share/applications too).  Defaults get
+      # set here and redeployed, not set at runtime.  `xdg-mime default`,
+      # `xdg-settings set` and Electron's `setAsDefaultProtocolClient` are
+      # shell writes that tmp-and-rename, so they replace the symlink and
+      # appear to work -- it is the *next* activation that would then fail, on
+      # an existing plain file it did not create, and `force` makes it
+      # overwrite instead.  (GIO's `g_app_info_set_as_default_for_type`, which
+      # is what Firefox's prompt uses, resolves the symlink first and so just
+      # fails against the read-only store.)
       mimeApps = {
         enable = true;
         defaultApplications = {
--- a/pkgs/launcher/home-manager-module.nix
+++ b/pkgs/launcher/home-manager-module.nix
@@ -28,5 +28,35 @@ in {
         lib.mapAttrsToList (name: path: {inherit name path;}) cfg.apps
       );
     };
+
+    # $BROWSER only reaches things that read it, which desktop apps mostly
+    # don't -- they resolve a web link through the xdg handler for the scheme
+    # and exec that .desktop's Exec line.  Without a default registered, that
+    # picks whichever installed app happens to claim the scheme; the ChatGPT
+    # app claims it for its in-app browser, so it was answering for every web
+    # link on the machine.  Point both routes at the same script instead.
+    xdg = {
+      desktopEntries.browse = {
+        name = "Browse";
+        genericName = "Web Browser";
+        exec = "${pkgs.launcher}/bin/browse %U";
+        terminal = false;
+        categories = ["Network" "WebBrowser"];
+        mimeType = [
+          "text/html"
+          "x-scheme-handler/http"
+          "x-scheme-handler/https"
+        ];
+      };
+
+      mimeApps = {
+        enable = true;
+        defaultApplications = {
+          "text/html" = "browse.desktop";
+          "x-scheme-handler/http" = "browse.desktop";
+          "x-scheme-handler/https" = "browse.desktop";
+        };
+      };
+    };
   };
 }
--- a/pkgs/launcher/scripts/browse.nix
+++ b/pkgs/launcher/scripts/browse.nix
@@ -37,11 +37,15 @@ writeShellScriptBin "browse" ''
     shift
   done
 
+  # Quoted, because this is the registered http/https handler now: a file://
+  # URL for a path with a space arrives as one argument and would otherwise
+  # split into two.  The :+ keeps an empty target from becoming an empty
+  # argument, which is what opens the browser on its homepage.
   case $browser in
-    brave) exec $brave $target >/dev/null 2>&1 ;;
-    chromium) exec $chromium $target >/dev/null 2>&1 ;;
-    tor-browser) exec $tor $target >/dev/null 2>&1 ;;
-    chrome) exec $chrome $target >/dev/null 2>&1 ;;
-    firefox) exec $firefox $target >/dev/null 2>&1 ;;
+    brave) exec $brave ''${target:+"$target"} >/dev/null 2>&1 ;;
+    chromium) exec $chromium ''${target:+"$target"} >/dev/null 2>&1 ;;
+    tor-browser) exec $tor ''${target:+"$target"} >/dev/null 2>&1 ;;
+    chrome) exec $chrome ''${target:+"$target"} >/dev/null 2>&1 ;;
+    firefox) exec $firefox ''${target:+"$target"} >/dev/null 2>&1 ;;
   esac
 ''


pkgs/launcher/home-manager-module.nix
```diff
@@ -28,5 +28,35 @@ in {
...
+        genericName = "Web Browser";
+        exec = "${pkgs.launcher}/bin/browse %U";
+        terminal = false;
+        categories = ["Network" "WebBrowser"];
+        mimeType = [
+          "text/html"
+          "x-scheme-handler/http"
+          "x-scheme-handler/https"
+        ];
+      };
```
cprussin: **Blocker — `browse.desktop` is never installed. This is the same `home.packages` `mkForce` trap you flagged for `xdg.portal.extraPortals` in the PR body.**

home-manager's `xdg.desktopEntries` does not write a file; it builds a `makeDesktopItem` derivation and adds it to `home.packages`. At the pinned rev (`flake.lock` → `home-manager` `65258d5c65a250189fde2e35f490d15e064c4c62`), `modules/misc/xdg-desktop-entries.nix`:

```nix
  config = lib.mkIf (config.xdg.desktopEntries != { }) {
    ...
    home.packages = (
      map lib.hiPrio # we need hiPrio to override existing entries
        (lib.attrsets.mapAttrsToList makeFile config.xdg.desktopEntries)
    );
  };
```

That is a normal-priority definition (100). `config/modules/ui/launcher/default.nix:42` sets `home.packages = lib.mkForce [pkgs.launcher];` (priority 50), as do ~17 other modules per the idiom documented in `config/modules/data/session-vars/home.nix:1-15`. The module system keeps only the highest-priority definitions, so every normal-priority contribution is dropped. `lib.hiPrio` does not help — it sets `meta.priority` on the derivation for `buildEnv` collision resolution, not a module-system override priority.

Concretely: `pkgs/launcher/home-manager-module.nix:20` (`home.packages = [pkgs.launcher];`) is already being discarded today — that's precisely why line 42 of the config module re-adds `pkgs.launcher` under `mkForce`. The new desktop item goes the same way.

The `mimeApps` half survives, because it routes through `xdg.configFile`/`xdg.dataFile` → `home.file`, which nothing here force-overrides. So the deployed result is:

- `~/.config/mimeapps.list` exists and names `browse.desktop` as the default for http/https/text/html,
- `browse.desktop` exists nowhere on `XDG_DATA_DIRS`,
- and ChatGPT no longer claims the schemes either.

Every web link resolves to nothing. That is worse than the status quo, and it will not show up in CI — `eval lyra` passes fine.

For the record, discoverability is otherwise sound, so this is the only thing standing between the change and working: `/etc/profiles/per-user/$USER` is on `environment.profiles` and `XDG_DATA_DIRS` is `profileRelative` `/share` (nixpkgs `d407951`, `nixos/modules/programs/environment.nix:38,51`), the per-user profile inherits `environment.pathsToLink` (`nixos/modules/config/users-groups.nix:970`), and `/share/applications` is not in the NixOS default list (`nixos/modules/config/system-path.nix:188-205`) but is added by `config/modules/ui/xdg-portal/default.nix:6`. Worth a word in the comment here that this module leans on `ui/xdg-portal` for that — they happen to be imported together in `config/profiles/pc/default.nix:34,44`, but nothing enforces it.

Two ways out. Either add the item to the `mkForce` list at `config/modules/ui/launcher/default.nix:42`, or — better, since it keeps the module self-contained and immune to the idiom — bypass `home.packages` entirely:

```nix
    xdg.dataFile."applications/browse.desktop".source =
      "${pkgs.makeDesktopItem {
        name = "browse";
        desktopName = "Browse";
        genericName = "Web Browser";
        exec = "${pkgs.launcher}/bin/browse %u";
        terminal = false;
        categories = ["Network" "WebBrowser"];
        mimeTypes = [
          "text/html"
          "x-scheme-handler/http"
          "x-scheme-handler/https"
        ];
      }}/share/applications/browse.desktop";
```

That lands in `~/.local/share/applications`, which the spec searches ahead of `XDG_DATA_DIRS`, so it also drops the dependency on `pathsToLink`.

---

pkgs/launcher/home-manager-module.nix
```diff
@@ -28,5 +28,35 @@ in {
...
+    # don't -- they resolve a web link through the xdg handler for the scheme
+    # and exec that .desktop's Exec line.  Without a default registered, that
+    # picks whichever installed app happens to claim the scheme; the ChatGPT
+    # app claims it for its in-app browser, so it was answering for every web
+    # link on the machine.  Point both routes at the same script instead.
+    xdg = {
+      desktopEntries.browse = {
+        name = "Browse";
+        genericName = "Web Browser";
+        exec = "${pkgs.launcher}/bin/browse %U";
```
cprussin: `%U` overpromises what `browse` does, and the script word-splits its target.

`pkgs/launcher/scripts/browse.nix:21-38`: the argument loop assigns `target=$1` on every non-flag argument, so with `%U`'s multiple URLs only the **last** one is opened and the rest are silently dropped. `%u` (single URL) is the honest field code here, and it's what the handler path actually delivers anyway.

Second, `browse.nix:41-45` runs `exec $brave $target` unquoted. Under `%U`/`%u` the desktop spec unescapes the argument before passing it, so a `file://` URL for a path with a space arrives as one argv element containing a space and then splits into two arguments to Brave. Quoting `"$target"` fixes it — though note the empty-target case (launched from a menu with no URL) currently relies on the unquoted expansion vanishing, so it'd need `''${target:+"$target"}` or an explicit branch.

Both are pre-existing in `browse`, but this PR is what puts it on the handler path for every web link on the machine, which is where they start to bite.

The rest of the `%U` edge cases check out: zero arguments leaves `target=""` and opens Brave on its homepage; a `file://` argument matches the `://` branch and is passed through unchanged; and `exec $brave "$target" >/dev/null 2>&1` behaves the same launched from a desktop `Exec` as from a terminal, since the redirections apply to the exec'd image and nothing here needs a tty.

---

pkgs/launcher/home-manager-module.nix
```diff
@@ -28,5 +28,35 @@ in {
...
+      };
+
+      mimeApps = {
+        enable = true;
+        defaultApplications = {
+          "text/html" = "browse.desktop";
+          "x-scheme-handler/http" = "browse.desktop";
+          "x-scheme-handler/https" = "browse.desktop";
+        };
+      };
```
cprussin, line 84: Non-blocking, but worth knowing you're signing up for it: `enable = true` makes `~/.config/mimeapps.list` a read-only store symlink (home-manager's own wording: *"The generated file is read-only"*), and it also writes the deprecated `~/.local/share/applications/mimeapps.list` alongside it.

From then on, anything that tries to *write* that file fails: `xdg-mime default ...`, `xdg-settings set default-web-browser ...`, Brave/Chromium's "make ChatGPT/Brave your default browser" prompt, and GTK/file-manager "Open With → always use for this type". That's arguably the point — one declarative source of truth — but it does mean the only way to change a default is to edit this file and redeploy, and the failures will be silent-ish permission errors rather than anything explanatory.

Nothing else in the repo manages `mimeapps.list` (only `config/modules/data/xdg-user-dirs/default.nix` and `config/modules/system/nix/default.nix:57` touch `xdg.*` at all), so there's no conflict to resolve — just a new constraint. A sentence in the module comment saying "defaults are set here, not with `xdg-mime`" would save future-you the debugging session.

---

pkgs/chatgpt-desktop/derivation.nix
```diff
@@ -212,10 +212,17 @@ in
...
+      # the handler for every web link, and the app resolves the login URL that
+      # way -- so clicking "sign in" re-execs the app, which hands off to the
+      # running instance ("Opening in existing browser session.") and exits,
+      # and no browser ever opens.  It should not be a general web browser.
       for desktop in "$out"/share/applications/*.desktop
       do
         substituteInPlace "$desktop" \
-          --replace-fail "Exec=chatgpt" "Exec=$out/bin/chatgpt"
+          --replace-fail "Exec=chatgpt" "Exec=$out/bin/chatgpt" \
+          --replace-fail "x-scheme-handler/http;x-scheme-handler/https;" ""
```
cprussin, line 232: This half is correct.

Against the shipped line

```
MimeType=x-scheme-handler/codex;x-scheme-handler/http;x-scheme-handler/https;text/csv;application/vnd.openxmlformats-...;...
```

removing `x-scheme-handler/http;x-scheme-handler/https;` leaves `MimeType=x-scheme-handler/codex;text/csv;...` — the two schemes sit in the middle of the list, both flanked by `;`, and taking the trailing `;` with them means no doubled separator, no leading `;`, no dangling one. `codex:` and the file-type associations survive, which is the right cut. And the `build pkgs.chatgpt-desktop` check is green on this commit, which is the real proof the literal matched — `--replace-fail` would have failed the build otherwise. Good call using `--replace-fail` rather than `--replace`; a `.deb` that reshuffles that list should break loudly.

One question about whether this reaches the machine at all, though. `pkgs.chatgpt-desktop` is never in `home.packages`, `environment.systemPackages`, or any profile — the only reference is `config/modules/ui/launcher/default.nix:57`, which puts the store path of the *binary* into the `.launcher-apps` linkFarm. So `$out/share/applications/chatgpt.desktop` isn't on `XDG_DATA_DIRS` and shouldn't be visible to the MIME database at all, yet your `strace` shows something resolving `https` to it.

The likely explanation is that the app self-registers at runtime — Electron apps routinely copy their bundled `.desktop` into `~/.local/share/applications` and run `update-desktop-database` — which would explain both the `Exec=` carrying an absolute store path (it copies the patched file) and how it got onto the handler list without being in a profile. If that's what's happening, then:

- a stale copy from a previous launch is sitting in `~/.local/share/applications` right now, still declaring `x-scheme-handler/http`,
- home-manager doesn't own that file, so redeploying will not remove or rewrite it,
- and it'll keep claiming the schemes until the app next overwrites it (or you delete it).

The `mimeApps` default would override it either way, since an explicit `[Default Applications]` entry beats a bare `MimeType=` claim — but only once `browse.desktop` actually exists (see the other comment). Worth adding to the verification steps: `grep -l x-scheme-handler/http ~/.local/share/applications/*.desktop`, remove any stale copy, `update-desktop-database ~/.local/share/applications`, then re-check `xdg-settings get default-web-browser`. And if the app *is* rewriting that file on every launch, the store-side patch is only as durable as the app's own copy of it, which is worth a line in the comment above.

---

cprussin (commented):
**Changes requested** (submitted as a comment — GitHub won't let a review request changes on its own author's PR).

The diagnosis is right and the ChatGPT half of the fix is correct. The second half doesn't work, for the exact reason you identified for `xdg.portal.extraPortals` in the "Noticed, not fixed here" section.

**`xdg.desktopEntries` installs through `home.packages`.** At the pinned home-manager rev, `modules/misc/xdg-desktop-entries.nix` builds a `makeDesktopItem` and adds it to `home.packages` at normal priority. `config/modules/ui/launcher/default.nix:42` sets `home.packages = lib.mkForce [pkgs.launcher];`, so it's discarded — the same trap, in the same module you're editing. (`lib.hiPrio` in that home-manager code is `meta.priority` for `buildEnv` collisions, not a module-system override.)

`xdg.mimeApps` is unaffected, because it goes through `xdg.configFile`/`xdg.dataFile` → `home.file`, which nothing here force-overrides. So the deployed machine gets a `mimeapps.list` naming `browse.desktop` as the default for http/https/text/html, no `browse.desktop` anywhere on `XDG_DATA_DIRS`, and a ChatGPT app that no longer claims the schemes either. Web links resolve to nothing — worse than today, and `eval lyra` won't catch it. Details and a suggested fix in the inline comment.

Everything else I checked holds up:

- The `--replace-fail` literal matches the shipped `MimeType=` exactly and leaves a well-formed line; the green `build pkgs.chatgpt-desktop` check proves the match.
- `browse.desktop` would be discoverable once actually installed — `/etc/profiles/per-user/$USER/share` is on `XDG_DATA_DIRS`, and `/share/applications` is in `pathsToLink` via `config/modules/ui/xdg-portal/default.nix:6` (it is *not* a NixOS default, so that's a real if currently-satisfied dependency).
- Nothing else in the repo manages `mimeapps.list`, and nothing else depends on ChatGPT handling http/https.
- `pkgs` and `lib` are both in scope at `pkgs/launcher/home-manager-module.nix:2-5`, `config` is still used, and `cli test lint`/`dead-code`/`format` are green — alejandra, statix and deadnix are satisfied.

Three smaller things inline: `%U` vs. what `browse` actually does with multiple URLs, unquoted `$target`, and the read-only `mimeapps.list` you're opting into. And one open question on whether the ChatGPT `.desktop` reaches the MIME database from the store at all, given the package isn't in any profile.

---

pkgs/launcher/home-manager-module.nix
```diff
@@ -28,5 +28,51 @@ in {
...
+      # browser's own "make me the default" prompt all stop working.  Defaults
+      # get set here and redeployed, not set at runtime.
+      mimeApps = {
+        enable = true;
+        defaultApplications = {
+          "text/html" = "browse.desktop";
+          "x-scheme-handler/http" = "browse.desktop";
+          "x-scheme-handler/https" = "browse.desktop";
+        };
+      };
```
cprussin: **Should fix before merge:** add `force = true` to the two `mimeapps.list` entries, or the first `nixos-rebuild switch` after this lands may abort.

`xdg.mimeApps.enable` writes *two* files ([`xdg-mime-apps.nix`](https://github.com/nix-community/home-manager/blob/65258d5c65a250189fde2e35f490d15e064c4c62/modules/misc/xdg-mime-apps.nix), pinned rev):

```nix
xdg.dataFile."applications/mimeapps.list".source = config.xdg.configFile."mimeapps.list".source;
xdg.configFile."mimeapps.list".source = ...;
```

Both route into `home.file`, and `home.activation.checkLinkTargets` runs `checkNewGenCollision` over **every** new-generation file on **every** activation. `check-link-targets.sh` errors when the target exists and isn't already an HM symlink:

```
collisionErrors+=("Existing file '$targetPath' would be clobbered")
```

which `exit 1`s the switch. Nothing in this repo sets `home-manager.backupFileExtension` or `backupCommand` (`config/modules/system/nix/default.nix:53-56` sets only `useGlobalPkgs`/`useUserPackages`), so there's no escape hatch — the deploy just fails until the file is deleted by hand.

`~/.config/mimeapps.list` is one of the most commonly pre-existing files in a home dir: anything that ever ran `xdg-mime default`, `xdg-settings set`, a GTK "Open With → set as default", or a browser's default-browser prompt creates it. On this machine specifically, `pkgs/chatgpt-desktop/derivation.nix:83` puts `xdg-utils` on the app's runtime PATH, and Electron's `setAsDefaultProtocolClient` is `xdg-settings set default-url-scheme-handler`, which writes exactly this file.

It's also worth being precise in the comment above: a runtime `xdg-mime default` / `xdg-settings set` / `g_app_info_set_as_default_for_type` doesn't cleanly "stop working". All three write a temp file and rename over the target, so they *replace* the read-only symlink with a real file and report success — and then the next deploy fails. With `force = true` the collision check is skipped and `ln -Tsf` just puts the managed file back, which is the behaviour the comment describes wanting. Home Manager's own error text recommends this exact option by name.

```nix
      mimeApps = {
        enable = true;
        defaultApplications = { ... };
      };

      # `force`, because the write path these tools use is tmpfile+rename:
      # they replace the symlink with a real file rather than failing against
      # the read-only store, and without this the next activation aborts on
      # the collision instead of just putting the managed file back.
      configFile."mimeapps.list".force = true;
      dataFile."applications/mimeapps.list".force = true;
```

---

pkgs/chatgpt-desktop/derivation.nix
```diff
@@ -212,10 +212,17 @@ in
...
       done
```
cprussin, line 233: Round 1's open question — "does the app self-copy its `.desktop` into `~/.local/share/applications`?" — I think the config answers it, and the answer is yes. Non-blocking, but it changes what this hunk is doing.

`pkgs.chatgpt-desktop` is never installed anywhere that has a `share/applications`:

- `config/modules/ui/launcher/default.nix:42` — `home.packages = lib.mkForce [pkgs.launcher]`, so the app isn't in the per-user profile;
- `config/modules/ui/launcher/default.nix:57` — `chatgpt = "${pkgs.chatgpt-desktop}/bin/chatgpt"`, a bare store path fed into the `.launcher-apps` linkFarm (`pkgs/launcher/home-manager-module.nix:26-29`), which is `~/.launcher-apps` and not on `XDG_DATA_DIRS`;
- `grep -rn systemPackages` finds no reference to it;
- `config/modules/ui/xdg-portal/default.nix:6` links `/share/applications` out of the *system* profile, which the app isn't in either.

So `$out/share/applications/chatgpt.desktop` is not discoverable by GIO. For the app to have been resolved as the `x-scheme-handler/http` handler at all, a copy must already be sitting in `~/.local/share/applications`, put there by the app itself at runtime. (It can't have come through `xdg-open` either — `pkgs/launcher/derivation.nix` shadows `xdg-open` with the launcher's `run`, which routes to `browse`.)

Two consequences:

1. This patch is not what fixes the reported symptom today. It only affects copies the app makes *after* a redeploy; the copy already on disk is unpatched and survives. The `mimeapps.list` half is the load-bearing one — `[Default Applications]` is consulted before any app that merely declares the MIME type, so it wins over the stale copy regardless. Worth saying so in the commit message rather than presenting the two halves as equally responsible.
2. A one-time `rm ~/.local/share/applications/chatgpt*.desktop` is still worth doing by hand, otherwise the stale entry keeps showing up in the "Open with" list for every web link.

It also feeds the `mimeapps.list` comment on the other file: an app that installs its own desktop entry generally follows up with `xdg-settings set default-url-scheme-handler`, which is precisely the write that will collide with the new read-only `~/.config/mimeapps.list`.

---

cprussin (commented):
Round 2 review of `43ddd13`. The round-1 blocker is genuinely fixed — I traced it rather than taking the comment's word for it. One new should-fix, one finding that settles round 1's open question, and everything else checks out.

## Round-1 blocker: fixed, verified

`xdg.dataFile` does survive the `mkForce` idiom. In home-manager at the pinned `65258d5c`, [`modules/misc/xdg.nix`](https://github.com/nix-community/home-manager/blob/65258d5c65a250189fde2e35f490d15e064c4c62/modules/misc/xdg.nix) routes it into `home.file` from an **unguarded** `mkMerge` arm — outside the `mkIf cfg.enable` block, so `xdg.enable` (never set in this repo) is irrelevant:

```nix
home.file = lib.mkMerge [
  ...
  (lib.mapAttrs' (name: file: lib.nameValuePair "${cfg.dataHome}/${name}" file) cfg.dataFile)
```

`home.stateVersion` is `26.05` (`config/modules/system/stateVersion/default.nix:2`), so the `versionAtLeast "20.09"` arm applies and `dataHome` is the deterministic `~/.local/share`, not the `getEnv` fallback. `grep -rn 'mkForce' --include=*.nix` over the repo shows 30 hits, all on `home.packages`, `systemPackages`, and unrelated scalars — nothing touches `home.file` or `xdg.dataFile`. And `files.nix` links per-file (`find "$newGenFiles" \( -type f -or -type l \)`), so `~/.local/share/applications` stays a real directory and other apps' entries there are undisturbed.

Also worth noting the reasoning in the comment holds for a second reason it doesn't mention: `useUserPackages = true` (`config/modules/system/nix/default.nix:53-56`) means `xdg.desktopEntries` *would* have landed somewhere on `XDG_DATA_DIRS` — via `/etc/profiles/per-user/...` — if it ever got there. It's specifically the `mkForce` that kills it, exactly as written.

## `makeDesktopItem`: correct

Checked against [the real nixpkgs rev this flake uses](https://github.com/NixOS/nixpkgs/blob/a9e6d84f9c2f9012f5fe7d964a7851352300e61a/pkgs/build-support/make-desktopitem/default.nix) — note that's `a9e6d84f`, the root `nixpkgs` input; `d4079514` is a transitive input (colmena's) and isn't what builds this.

- Every argument name is real and correctly spelled: `name`, `desktopName`, `genericName`, `exec`, `terminal`, `categories`, `mimeTypes`. No typo'd key silently dropped (the function is a non-`@args` attrset pattern, so a typo would have been an eval error anyway).
- Output path: `writeTextFile { destination = "${destination}/${name}.${extension}"; }` with `destination ? "/share/applications"` and `extension = "desktop"` for `type = "Application"` ⇒ `$out/share/applications/browse.desktop`. It names the file after `name`, *not* `desktopName`, so `name = "browse"` / `desktopName = "Browse"` gives the lowercase basename the interpolation expects. Good — swapping those two would have produced `Browse.desktop` and a dangling `.source`.
- Generated content is a valid entry: `[Desktop Entry]`, `Type=Application`, `Version=1.5`, `Name=Browse`, plus `Exec`/`Terminal`/`MimeType`/`Categories`. `checkPhase` runs `desktop-file-validate` on it, so an invalid entry fails the build rather than shipping.
- `terminal = false` renders `Terminal=false` via `boolOrNullToString`, not `Terminal=0` or a Nix-y `false`.

## Basename / MIME interaction: correct

`xdg.dataFile."applications/browse.desktop"` ⇒ `~/.local/share/applications/browse.desktop` ⇒ desktop id `browse.desktop`, which is what the three `defaultApplications` entries name. And the entry's own `MimeType=` is irrelevant to the `[Default Applications]` lookup — GIO resolves the id straight to a `GDesktopAppInfo` without checking that the app declares the type, and `xdg-mime query default` doesn't check either. It only matters for the "Open with…" enumeration, where it's also correct. So the two halves line up.

## `~/.local/share/applications` precedence: yes, and it is searched

Per the basedir spec `$XDG_DATA_HOME` (default `~/.local/share`) precedes `$XDG_DATA_DIRS`. `grep -rn 'XDG_DATA_HOME'` over the repo returns nothing, and home-manager only exports it under `xdg.enable`, which is unset — so the spec default applies and no unusual value is in play. No `update-desktop-database` / `mimeinfo.cache` dependency for the default-app path.

## `''${target:+"$target"}`: correct in all four cases

`''${` is the Nix indented-string escape for a literal `${`, so the generated script contains `${target:+"$target"}`, not `''$...`. Behaviour:

- **empty** — `:+` yields nothing, zero words, `exec $brave` opens the homepage. Same as before, which is the intent.
- **normal URL** — one quoted word.
- **`file://` with a space** — the inner `"$target"` *is* honoured; within `${p:+word}` the quote characters in `word` are processed as quoting, so no field splitting. This is the standard `${1:+"$1"}` idiom and it's used correctly.
- **`$` or backtick in the target** — parameter expansion doesn't recurse, so nothing re-expands.

It's also a strict improvement over the old unquoted `$target`, which would have globbed a URL containing `*` or `?` against the cwd.

## No regression for existing callers

`brave`/`chromium`/`firefox`/`chrome`/`tor-browser` in `config/modules/ui/launcher/default.nix:52-98` all call `browse --browser X $*`; the flag parsing is untouched and the empty-target path is byte-identical in behaviour. `$BROWSER` (`pkgs/launcher/home-manager-module.nix:24`) still points at the same script. `run`'s `exec $browse "$1"` passes one already-quoted argument, unaffected.

## Nix style

No findings — `cli test` (lint / dead-code / format) is green on `43ddd13`, as are `flake check` and `eval lyra`/`eval crux`, so alejandra, statix and deadnix are all satisfied and the module evaluates. `build pkgs.chatgpt-desktop` is green too, which incidentally proves the new `--replace-fail "x-scheme-handler/http;x-scheme-handler/https;"` matches in *every* `.desktop` the `.deb` ships — that loop would have failed the build otherwise.

## What's left

Two inline comments: one should-fix (`force = true` on the `mimeapps.list` entries, or the first deploy can abort on a collision), and one finding that settles round 1's open question about the self-copy. Neither is a correctness problem with the code as written; the first is a deploy-robustness gap and the second is mostly about what the commit message claims. Happy to approve once the `force` one is addressed.

---

cprussin:
## Round 3 review — head `b14c003`

**Verdict: approve.** Both round-2 findings are addressed and hold up under independent checking. One non-blocking correction to a comment, spelled out below; the code is right as it stands.

(Posted as an issue comment because the review API is rate-limited on this account right now — the one inline point is anchored below with file:line.)

### Round-2 fix 1 — `force = true` on both `mimeapps.list` paths: verified correct

Checked against home-manager `65258d5c65a250189fde2e35f490d15e064c4c62` (root `flake.lock` → `home-manager`; root nixpkgs is `a9e6d84f9c2f9012f5fe7d964a7851352300e61a`, and `d4079514` is colmena's transitive input, so I verified `makeDesktopItem` against `a9e6d84f`).

- `xdg.configFile` / `xdg.dataFile` both use `fileType` from `modules/lib/file-type.nix`, which defines `force` (line 133, `default = false`) alongside `source`/`target`. `modules/misc/xdg.nix` maps every one of them into `home.file` verbatim via `lib.mapAttrs' (name: file: nameValuePair "${cfg.dataHome}/${name}" file)` — the whole submodule value carries over, `force` included.
- `modules/files.nix:146-148` collects `forcedPaths` as `map (v: v.target) (filter (v: v.force) cfg)`. `target` is normalised by `file-type.nix:47-51` (`removePrefix (homeDirectory + "/")`), so `xdg.dataFile."applications/mimeapps.list"` becomes `.local/share/applications/mimeapps.list` and lands in `forcedPaths` correctly. The absolute-path round-trip through `home.file` is idempotent.
- `modules/files/check-link-targets.sh` skips the collision check for any path under a forced prefix (`verboseEcho "Skipping collision check for $targetPath"`), so `checkNewGenCollision` no longer `exit 1`s. The link phase then runs `ln -Tsf` (`files.nix:215-218`), which replaces a plain file. So the abort is genuinely gone, not just deferred.
- Nothing over-reaches. The prefix match (`$targetPath == $forcedPath*`) could in principle swallow a sibling like `mimeapps.list.bak`, but no such HM-managed file exists here. And `grep -rn "mimeApps\|mimeapps\|xdg.enable\|desktopEntries\|dataHome"` across the repo returns only this module, so there is no competing definition.

Worth knowing (not a change request): `force` is documented as *"will silently delete the target regardless of whether it is a file or link"*, and with no `backupFileExtension` set, a pre-existing `~/.config/mimeapps.list` is destroyed without a copy on first activation — including any unrelated defaults in it (PDF viewer, image viewer). That is the intended trade for a declarative source of truth, and it is what home-manager's own error text recommends by name, so I'd leave it. Just don't be surprised.

### Round-2 fix 2 — the derivation comment: verified accurate

`grep -rn chatgpt-desktop --include=*.nix` finds exactly three hits, none of them a profile: the overlay import, `config/modules/ui/launcher/default.nix:57` (`chatgpt = "${pkgs.chatgpt-desktop}/bin/chatgpt"`, a bare store path into the `.launcher-apps` linkFarm), and a zfs mountpoint in `config/machines/lyra/default.nix:111`. So `$out/share/applications` is never on `XDG_DATA_DIRS`, and the comment at `pkgs/chatgpt-desktop/derivation.nix:196-207` describes the situation correctly.

### Round-1/2 conclusions re-verified on this head

- **`xdg.dataFile` survives the `mkForce` idiom.** `modules/misc/xdg-desktop-entries.nix:237-240` really does route `xdg.desktopEntries` through `home.packages`, and `config/modules/ui/launcher/default.nix:42` is `home.packages = lib.mkForce [pkgs.launcher];`, so the old approach would have been dropped. `xdg.dataFile` → `home.file` bypasses that entirely. Correct call.
- **`browse.desktop` basename.** nixpkgs `a9e6d84f`, `pkgs/build-support/make-desktopitem/default.nix:198-206`: `destination` defaults to `/share/applications` and the file is `${name}.${extension}` with `extension = "desktop"`. `.../share/applications/browse.desktop` (`pkgs/launcher/home-manager-module.nix:65`) is right. The builder also runs `desktop-file-validate` in `checkPhase`, and `Categories=Network;WebBrowser;` plus the `MimeType` list are valid, so that won't bite at build time.
- **`''${target:+"$target"}`** renders as `${target:+"$target"}` — `''${` is the Nix indented-string escape for a literal `${`. Correct in all five branches (`pkgs/launcher/scripts/browse.nix:45-49`).
- **Nix style.** CI is green on `b14c003`, including `cli test` (which runs `lint` / `dead-code` / `format`, i.e. statix / deadnix / alejandra) and `flake check --all-systems`, plus `eval lyra` and `eval crux`. Nothing to add by eye either.

---

### Non-blocking: the tmp+rename claim is right for `xdg-mime`/`xdg-settings`, wrong for GIO

**`pkgs/launcher/home-manager-module.nix:69-73`**

```nix
      # set here and redeployed, not set at runtime -- but `xdg-mime default`,
      # `xdg-settings set` and a browser's "make me the default" prompt all
      # write tmp-and-rename, so they replace the symlink and appear to work.
```

The round-2 comment asserted this of all three write paths, including `g_app_info_set_as_default_for_type`, and this commit carried that into the code. GLib is the exception, and it is the one that matters for "a browser's prompt".

`gio/gdesktopappinfo.c`, `update_mimeapps_list()` — it resolves the symlink *before* writing:

```c
  filename = g_build_filename (dirname, "mimeapps.list", NULL);

  while (g_file_test (filename, G_FILE_TEST_IS_SYMLINK))
    {
      old_filename = filename;
      filename = g_file_read_link (old_filename, error);
      ...
    }
  ...
  res = g_file_set_contents_full (filename, data, (gssize) data_size,
                                  G_FILE_SET_CONTENTS_CONSISTENT | G_FILE_SET_CONTENTS_ONLY_EXISTING,
                                  0600, error);
```

So with `~/.config/mimeapps.list` a store symlink, `filename` becomes the `/nix/store/...-mimeapps.list` path, and the `G_FILE_SET_CONTENTS_CONSISTENT` tmp+rename is attempted *inside the store directory* — which fails. It does **not** replace the symlink, it does **not** appear to work, and it does **not** break the next activation. It just returns `FALSE` with a `GError`.

Concretely that splits the browsers:

- Chrome / Chromium / Brave "make me default" → `xdg-settings set default-browser` → the `xdg-utils` shell path (`awk … > "$file.new"; mv`), which *does* replace the symlink. Same for Electron's `setAsDefaultProtocolClient`, which is `xdg-settings set default-url-scheme-handler` — the ChatGPT app's own write.
- Firefox "make me default" → `nsGNOMEShellService` → `g_app_info_set_as_default_for_type` → the GLib path above, which fails outright.

**This does not change the fix.** `force = true` is still right and still needed, for two independent reasons: the first-deploy-onto-an-existing-plain-file case (which was round 2's actual blocker and has nothing to do with runtime writes), and the `xdg-mime`/`xdg-settings`/Electron path, which really does leave a plain file behind. Only the comment's blanket "all three" is wrong.

Suggested wording for lines 69-73:

```nix
      # set here and redeployed, not set at runtime.  `xdg-mime default`,
      # `xdg-settings set` and Electron's `setAsDefaultProtocolClient` are
      # shell writes that tmp-and-rename, so they replace the symlink and
      # appear to work -- it is the *next* activation that would then fail, on
      # an existing plain file it did not create, and `force` makes it
      # overwrite instead.  (GIO's `g_app_info_set_as_default_for_type`, which
      # is what Firefox's prompt uses, resolves the symlink first and so just
      # fails against the read-only store.)
```

Fine to fold in on the next push or to leave; it is a comment, and the behaviour it guards is correct either way. Nothing else outstanding — recommend merge once this is settled.

---

PR #22, opened 2026-08-28
[xdg-portal] Actually install the portal backends

cprussin:
`xdg.portal.extraPortals` has been installing nothing on this machine.

## The bug

home-manager's portal module installs the frontend and every entry of `extraPortals` through `home.packages` (`modules/misc/xdg-portal.nix:172` at the pinned rev):

```nix
home = {
  packages = packages ++ cfg.configPackages;
```

That's a normal-priority definition, and this repo's `lib.mkForce` on `home.packages` — the idiom documented in `config/modules/data/session-vars/home.nix:1-15`, used in ~18 modules — discards every normal-priority contribution. So `extraPortals = [wlr gtk]` at `config/modules/ui/sway/default.nix:205` installed neither.

The only reason anything works today is that the sway module happens to name `pkgs.xdg-desktop-portal-wlr` by hand in its own `mkForce` list (`sway/default.nix:24`). Confirmed on the session bus:

```
$ busctl --user list | grep -i portal
org.freedesktop.impl.portal.desktop.wlr    3864  ...
org.freedesktop.portal.Desktop             3838  ...
```

No `gtk`. And `wlr` implements only Screenshot and ScreenCast, so FileChooser, AppChooser, Settings, Print and the rest have had **no implementation at all** — including the AppChooser the frontend's own `OpenURI` falls back to when a URL has no default handler. That's what these are, logged by every GTK app on startup:

```
org.freedesktop.DBus.Error.InvalidArgs: No such interface "org.freedesktop.portal.FileChooser"
org.freedesktop.DBus.Error.UnknownMethod: No such interface "org.freedesktop.portal.Settings"
```

## The fix

**gtk gets named in a `mkForce` list**, the same way wlr already was, with a comment at both sites explaining why setting the option alone isn't enough. `extraPortals` stays set — home-manager asserts on an empty list, and it's what the generated `portals.conf` is validated against — but the package list is what installs.

Naming the package is the whole fix. Every other path a backend needs is already in place:

- `/share/dbus-1`, which carries the service file the frontend activates the backend by name through, comes from `nixos/modules/services/system/dbus.nix:123-126` (this repo enables `services.dbus`).
- `/share/systemd`, which carries the unit that service file names, is a `pathsToLink` default (`nixos/modules/config/system-path.nix:203`), inherited by the per-user profile (`users-groups.nix:990-996`) and reached via `XDG_DATA_DIRS`.

That second one is demonstrated by the frontend itself: `xdg-desktop-portal` also carries `SystemdService=`, has no `[Install]` section and no socket unit, is in no `systemd.packages` anywhere in this config, and is running today — visible in the `busctl` output above. The profile route is the only way it can be starting.

**sway's `configPackages` comes off.** It goes the same way as `extraPortals`, and the config it would have supplied is set explicitly on the next line, so nothing depended on that package reaching the profile.

## Corrected from earlier revisions of this PR

Two claims I made and then withdrew, recorded so the history reads straight:

- An earlier revision added `/share/dbus-1` to `environment.pathsToLink`, claiming the service files were being dropped. Wrong — `services.dbus` already contributes it. Removed.
- A later revision then claimed backends install units to `$out/lib/systemd/user`, outside systemd's search path, and added `systemd.packages` to compensate. Also wrong — `move-systemd-user-units.sh` is a default stdenv fixup hook (`pkgs/stdenv/generic/default.nix:81`) that relocates them to `$out/share/systemd/user`, and `/share/systemd` is a `pathsToLink` default. That made the `systemd.packages` line redundant; removed.

## Relationship to #21

Found while diagnosing the ChatGPT sign-in bug, but **not** its cause — `dbus-monitor` during a sign-in click shows no `OpenURI` call at all. Independent bug, independent fix.

## Verification

CI covers evaluation. The behavioural check needs hardware — after deploying:

```
busctl --user list | grep -i portal
```

should now show `org.freedesktop.impl.portal.desktop.gtk` alongside `wlr`, and the `No such interface` errors should stop appearing in app logs.

https://claude.ai/code/session_013VHH9pj1eDhtmmfNwE4ZnY

### DIFF
--- a/config/modules/ui/sway/default.nix
+++ b/config/modules/ui/sway/default.nix
@@ -201,9 +201,16 @@ in {
       '';
     };
 
+    # wlr implements Screenshot and ScreenCast; everything else a portal is
+    # asked for comes from gtk, which ui/xdg-portal contributes.  Naming the
+    # package in `home.packages` above as well is what actually installs it --
+    # see the comment there for why `extraPortals` on its own doesn't.
+    #
+    # `configPackages` goes the same way and is left off deliberately: the
+    # config below is set explicitly rather than read out of sway's own
+    # portals.conf, so nothing depends on that package reaching the profile.
     xdg.portal = {
-      extraPortals = [pkgs.xdg-desktop-portal-wlr pkgs.xdg-desktop-portal-gtk];
-      configPackages = [pkgs.sway];
+      extraPortals = [pkgs.xdg-desktop-portal-wlr];
       config.sway.default = ["wlr" "gtk"];
     };
   };
--- a/config/modules/ui/xdg-portal/default.nix
+++ b/config/modules/ui/xdg-portal/default.nix
@@ -3,10 +3,39 @@
   lib,
   ...
 }: {
-  environment.pathsToLink = ["/share/xdg-desktop-portal" "/share/applications"];
+  # Neither is a NixOS default: /share/xdg-desktop-portal carries the .portal
+  # files the frontend reads to discover backends, /share/applications the
+  # desktop entries.  With `useUserPackages` the per-user profile is filtered
+  # by this list, so a backend package that isn't linked here is invisible.
+  # (/share/dbus-1, which carries the service files the frontend activates a
+  # backend by name through, is already linked by `services.dbus`; /share/
+  # systemd, which carries the units those service files name, is a default.)
+  environment.pathsToLink = [
+    "/share/applications"
+    "/share/xdg-desktop-portal"
+  ];
 
   primary-user.home-manager = {
-    home.packages = lib.mkForce [pkgs.xdg-desktop-portal];
-    xdg.portal.enable = true;
+    # home-manager installs the frontend and everything in `extraPortals`
+    # through `home.packages`, at the normal priority this repo's `lib.mkForce`
+    # on that list discards -- so `extraPortals` alone installs nothing, and
+    # each backend has to be named here too.  Setting it is still required
+    # (home-manager asserts on an empty list) and is what the generated
+    # portals.conf is checked against.
+    home.packages = lib.mkForce [
+      pkgs.xdg-desktop-portal
+      pkgs.xdg-desktop-portal-gtk
+    ];
+
+    # gtk is the backend that implements FileChooser, AppChooser, Settings,
+    # Print and most of the rest.  wlr, added by the sway module, implements
+    # only Screenshot and ScreenCast, so without this the interfaces every GTK
+    # app asks for on startup have no implementation at all -- including the
+    # AppChooser that the frontend's own OpenURI falls back to when there is no
+    # default handler for a URL.
+    xdg.portal = {
+      enable = true;
+      extraPortals = [pkgs.xdg-desktop-portal-gtk];
+    };
   };
 }


cprussin:
## Review

The core fix is right and I verified it end-to-end against the pinned inputs (root `nixpkgs` = `a9e6d84f9c2f9012f5fe7d964a7851352300e61a`, home-manager = `65258d5c65a250189fde2e35f490d15e064c4c62`). Two of the three explanations attached to it are factually wrong, though, and one of the three code changes is a no-op that only exists because of one of those wrong explanations.

Posting as a single comment rather than inline threads — the review API is rate-limited right now. File:line citations below.

### What checks out

- **The diagnosis of the actual bug is correct.** `modules/misc/xdg-portal.nix` at the pinned HM rev does `home.packages = packages ++ cfg.configPackages;` where `packages = [pkgs.xdg-desktop-portal] ++ cfg.extraPortals`. Normal priority, so the repo's `lib.mkForce` idiom eats it. Naming gtk in the `mkForce` list is the right fix.
- **`extraPortals` merging is fine.** `config/modules/ui/xdg-portal/default.nix:36` and `config/modules/ui/sway/default.nix:213` are two equal-priority (mkForce is not involved here) list definitions of the same option, so the effective value is `[wlr gtk]` and HM's `assertion = cfg.extraPortals != []` holds. Both `eval crux` and `eval lyra` are green, which confirms it — that assertion and the `pathsToLink` one (`/share/applications` + `/share/xdg-desktop-portal`, both still present) are evaluated there.
- **Dropping `configPackages` loses nothing — I checked rather than assumed.** In the pinned HM rev `cfg.configPackages` is referenced exactly once in the whole module, in that same `home.packages` line. Nothing turns it into a config dir; the option's docstring name-drops `NIXOS_XDG_DESKTOP_PORTAL_CONFIG_DIR` but no code sets it. It was `mkForce`d away anyway. And the `warnings = optional (cfg.configPackages == [] && cfg.config == {})` guard still doesn't fire, because `config.sway` is set. So `config/modules/ui/sway/default.nix:206` was inert in both directions. Good removal.
- **No screen-sharing regression.** `gtk.portal`'s `Interfaces=` (upstream `data/gtk.portal.in`, 1.15.3) lists FileChooser, AppChooser, Print, Notification, Inhibit, Access, Account, Email, DynamicLauncher, Lockdown, Settings, Wallpaper — no Screenshot, no ScreenCast. So `default = ["wlr" "gtk"]` cannot route either of those to gtk even in principle. `wlr.portal` still claims both. Nothing gets worse.
- **`UseIn=gnome` on `gtk.portal` is a non-issue** — with a matching `sway-portals.conf` present, xdp 1.20 never falls back to `UseIn`.
- Style is clean: alejandra-formatted, `lib` and `pkgs` both used in both files, nothing for statix/deadnix. `cli test` is green.

---

### 1. `/share/dbus-1` is already in `environment.pathsToLink` — the addition is a no-op and the comment is wrong

`config/modules/ui/xdg-portal/default.nix:9-11`:

> `None of the three is a NixOS default`

Two of the three aren't. `/share/dbus-1` is. `nixos/modules/services/system/dbus.nix:123-126` at the pinned rev:

```nix
config = mkIf cfg.enable (mkMerge [
  {
    environment.etc."dbus-1".source = configDir;

    environment.pathsToLink = [
      "/etc/dbus-1"
      "/share/dbus-1"
    ];
```

and `config/modules/system/dbus/default.nix:2-3` sets `services.dbus.enable = true`, so that branch is live on every machine in this repo. The PR body cites `nixos/modules/config/system-path.nix:188-205` for the claim — that's the right file for `/share/systemd` and friends, but it isn't the only module that contributes to `pathsToLink`, and dbus's own module is the one that covers its own directory.

The knock-on is that the stated failure mechanism is also wrong: the service files were **not** being dropped. `share/dbus-1/services/org.freedesktop.impl.portal.desktop.gtk.service` would have been linked into the per-user profile all along (`nixos/modules/config/users-groups.nix:990-994` builds it with `inherit (config.environment) pathsToLink`). gtk was missing for exactly one reason — the package was never installed. Which is the reason the `mkForce` half of this PR is both correct and sufficient.

Worth confirming while we're here, since it was the thing that would have made this dangerous: `pathsToLink` also feeds `system.path`, but `nixos/modules/config/system-path.nix:207-211` sets `ignoreCollisions = true`, so there's no system-wide dbus service collision risk from adding it. It's harmless, just pointless.

Suggest either dropping `"/share/dbus-1"` from the list and the clause from the comment, or keeping the line and rewording to something honest — e.g. "`services.dbus` already adds `/share/dbus-1`; restated here so this module doesn't silently depend on that."

### 2. The `wlr survived only because /share/systemd is a NixOS default` claim doesn't hold

This one is in the commit message and PR body rather than in-tree, but it's load-bearing for the reader and it's wrong twice.

Portal backends do not install their systemd user unit under `share/systemd`. xdg-desktop-portal-gtk's `meson.build:20-24` is explicit about it:

```
systemd_userunit_dir = get_option('systemd-user-unit-dir')
if systemd_userunit_dir == ''
  # This is deliberately not ${libdir}: systemd units always go in
  systemd_userunit_dir = prefix / 'lib' / 'systemd' / 'user'
```

wlr's `meson.build:92-103` takes it from `systemd.pc`'s `systemduserunitdir` with `pkgconfig_define: ['prefix', get_option('prefix')]`, which resolves to the same `$out/lib/systemd/user`. `/lib` happens to be in the default `pathsToLink` too, so both units do reach the profile — but at `<profile>/lib/systemd/user/`, which is not on systemd's user unit search path (that's `$XDG_DATA_DIRS/systemd/user`, i.e. `share/`). The NixOS `xdg.portal` module papers over this with `systemd.packages = packages` (`nixos/modules/config/xdg/portal.nix:139`), which `generateUnits` turns into `/etc/systemd/user` entries by globbing `$i/lib/systemd/${typeDir}/*` (`nixos/lib/systemd-lib.nix:444`). Home-manager's portal module does none of that.

So what actually put wlr on the bus is its `share/dbus-1/services/…` file plus D-Bus activation — the thing that was already working. Please reword; as written it will send the next reader looking at the wrong directory.

### 3. One residual risk worth a `busctl` check after deploy (not a blocker)

Following from the above: gtk's D-Bus service file carries `SystemdService=` unconditionally (upstream `data/org.freedesktop.impl.portal.desktop.gtk.service.in`):

```
[D-BUS Service]
Name=org.freedesktop.impl.portal.desktop.gtk
Exec=@libexecdir@/xdg-desktop-portal-gtk
SystemdService=xdg-desktop-portal-gtk.service
```

wlr's is templated — `conf_data.set('systemd_service', ...)` only when `dependency('systemd', required: get_option('systemd'))` is found, and that option defaults to `auto` (`meson_options.txt`), with nixpkgs passing no `-Dsystemd=` flag. So *if* wlr's build happened not to find `systemd.pc`, wlr activates purely through `Exec=` while gtk gets routed to `systemd --user` for a unit that isn't in its search path. I believe systemdLibs does ship `systemd.pc` (the `devel` install tag is retained under `buildLibsOnly`, `pkgs/os-specific/linux/systemd/default.nix:607-611`), which would make the two symmetric and gtk a non-event — but I can't build here to confirm, and the PR's own verification step covers it anyway:

```
busctl --user list | grep -i portal
```

If `org.freedesktop.impl.portal.desktop.gtk` doesn't appear after deploying, the one-line fix is `systemd.packages = [pkgs.xdg-desktop-portal-gtk];` at the NixOS level, matching what the NixOS portal module does. Might be worth adding pre-emptively — it's cheap and it makes the config independent of a build-time `auto` feature detection in a third-party package. Your call.

---

### Verdict

Requesting changes on the comment/commit-message accuracy in (1) and (2) only. The three code edits are correct: gtk in the `mkForce` list is the real fix, the `configPackages` removal is verifiably inert, and the `/share/dbus-1` line is harmless-but-redundant. Nothing here makes the current state worse.

---

cprussin:
## Review — round 2 (posting as a comment; review API is rate-limited)

**Requesting changes.**

The core fix (naming the backends in the `mkForce` `home.packages` lists) is right and the `/share/dbus-1` removal is correctly reasoned. But the round-1 "correction" about `/share/systemd` was itself wrong, and both the new `systemd.packages` comment and the wlr comment are built on it. Two factual blockers.

Verified against root nixpkgs `a9e6d84f9c2f9012f5fe7d964a7851352300e61a` (confirmed root: `nixpkgs_6` in `flake.lock`; `d4079514` is colmena's) and home-manager `65258d5c`.

---

### 1. Blocker — `config/modules/ui/sway/default.nix:208-210` is factually wrong

> its D-Bus service file only names a SystemdService when its build finds systemd, so it activates through plain `Exec=`

wlr's build **does** find systemd, so the service file **does** carry `SystemdService=`.

`meson.build` at `v0.8.2` (the pinned version) installs the unit *only* inside the `if systemd.found()` block, and that same block is what sets the `SystemdService=` line:

```meson
systemd = dependency('systemd', required: get_option('systemd'))   # option default: auto
if systemd.found()
    conf_data.set('systemd_service', 'SystemdService=' + systemd_service_file)
    configure_file(... install_dir: user_unit_dir)
endif
```

The built output contains the unit, so `systemd.found()` was true. From `cache.nixos.org`'s file listing for `xdg-desktop-portal-wlr`:

```
symlink /lib/systemd/user -> ../../share/systemd/user
regular /share/dbus-1/services/org.freedesktop.impl.portal.desktop.wlr.service
regular /share/systemd/user/xdg-desktop-portal-wlr.service
regular /share/xdg-desktop-portal/portals/wlr.portal
```

Mechanism: `buildInputs` has `systemdLibs`, whose `mesonInstallTags` includes `"devel"`; systemd's `src/core/meson.build` tags `systemd.pc` with `install_tag : 'devel'`, and nixpkgs points `pkgconfigdatadir` at `${placeholder "dev"}/share/pkgconfig`. So `dependency('systemd')` resolves. (`pkgs/by-name/xd/xdg-desktop-portal-wlr/package.nix` is unchanged between the pinned 0.8.2 and master's 0.8.4 apart from version/hash, so that listing is representative.)

wlr is in exactly the same position as gtk, not a counter-example to it.

### 2. Blocker — `config/modules/ui/xdg-portal/default.nix:17-18` is factually wrong

> Backends install their systemd user unit to `$out/lib/systemd/user`, which is not on systemd's search path.

They don't end up there, and `/share/systemd` **is** a NixOS `pathsToLink` default. Round 1's original claim was correct; this "correction" reintroduced the error.

* `pkgs/build-support/setup-hooks/move-systemd-user-units.sh` (a default fixup hook) moves `$out/lib/systemd/user/*` into `$out/share/systemd/user/` and replaces the old dir with a symlink. Confirmed in all three store listings — gtk, wlr and the frontend each ship `share/systemd/user/*.service` with `lib/systemd/user` as a compat symlink.
* `nixos/modules/config/system-path.nix:188-205` — the default `environment.pathsToLink` includes `"/share/systemd"`.
* `nixos/modules/config/users-groups.nix:994` — the per-user profile at `/etc/profiles/per-user/$USER` is built with `inherit (config.environment) pathsToLink`, so that default applies there too.
* `nixos/modules/programs/environment.nix` — `environment.profileRelativeSessionVariables.XDG_DATA_DIRS = [ "/share" ]`, and `$XDG_DATA_DIRS/systemd/user` is on systemd's user unit search path (see `systemd_user_unit_path` in `systemd.pc`).

So a backend in `home.packages` already lands somewhere `systemd --user` looks.

**Direct proof:** `xdg-desktop-portal` (the frontend) has `SystemdService=xdg-desktop-portal.service` in `org.freedesktop.portal.Desktop.service` at 1.20.4, ships its unit only under `share/systemd/user/`, and is **not** in `systemd.packages` in this repo — yet it is running today (it's in your own `busctl` output in the PR body). That can only work via the profile route.

### 3. Consequence — the `systemd.packages` line (`config/modules/ui/xdg-portal/default.nix:22`)

To answer the question directly: **it is not a no-op.** `nixos/modules/system/boot/systemd/user.nix` does

```nix
environment.etc."systemd/user".source = generateUnits { type = "user"; inherit (cfg) units; ... };
```

and `generateUnits` defaults `packages ? cfg.packages` (`nixos/lib/systemd-lib.nix:385`), globbing `$pkg/etc/systemd/user/*` and `$pkg/lib/systemd/user/*` — the latter resolving through the compat symlink. So the unit really does land in `/etc/systemd/user`, which is unconditionally on the search path.

It's also **safe**: `generateUnits` skips `.wants` dirs, so nothing is enabled or ordered into a target — no start at the wrong time. Both copies are symlinks to the same store file, so no double-start and no D-Bus/systemd activation conflict; `SystemdService=` stays the single start path.

It is just **redundant**, and the comment justifying it is wrong. Two coherent ways out:

* **Drop `:17-22`.** Simplest, and consistent with how the frontend and wlr already reach the bus.
* **Keep it and say what it actually does** — an extra, higher-priority copy at `/etc/systemd/user`, mirroring NixOS' module. If you keep it, make it symmetric: `nixos/modules/config/xdg/portal.nix:139` is `systemd.packages = packages` where `packages = [ pkgs.xdg-desktop-portal ] ++ cfg.extraPortals`, i.e. the frontend and *every* backend. Naming gtk alone doesn't match it.

Either way `sway/default.nix:208-210` needs rewriting, and the PR body's "portal backends don't install units under `share/`" needs to come out.

---

### Re-verified from round 1 — still good, no changes wanted

* **`/share/dbus-1` removal** — correct. `nixos/modules/services/system/dbus.nix:123-126` contributes `["/etc/dbus-1" "/share/dbus-1"]` under `mkIf cfg.enable`, and `config/modules/system/dbus/default.nix:2` sets `services.dbus.enable = true` in the same `pc` profile that imports this module. The comment at `xdg-portal/default.nix:10-11` is accurate.
* **`xdg-portal/default.nix:6-9`** — accurate. Neither `/share/applications` nor `/share/xdg-desktop-portal` is in the `system-path.nix` defaults, and home-manager's own assertion (`modules/misc/xdg-portal.nix`) demands exactly those two under `useUserPackages`, which `config/modules/system/nix/default.nix:55` enables.
* **gtk "unconditionally"** — accurate. `data/org.freedesktop.impl.portal.desktop.gtk.service.in` at 1.15.3 hardcodes `SystemdService=xdg-desktop-portal-gtk.service`, and `data/meson.build` installs both files unconditionally.
* **`extraPortals` merge** — `[gtk]` here plus `[wlr]` from sway merges to a two-element list, so the HM assertion (`extraPortals != []`) holds. Non-sway `pc` hosts are actually better off than before, when only sway supplied a backend.
* **`configPackages` removal is inert** — at this HM rev `configPackages` feeds only `home.packages` (discarded by the `mkForce`) and suppresses a warning that `xdg.portal.config.sway.default` already suppresses. The comment at `sway/default.nix:212-214` is accurate.
* **No ScreenCast/Screenshot regression** — wlr stays in both `home.packages` (`sway/default.nix:24`) and `extraPortals`, and `config.sway.default = ["wlr" "gtk"]` is unchanged.
* **Style** — `pkgs` and `lib` are both in scope and used in both files; formatting matches alejandra, and CI (`flake check`, both `eval` jobs) is green on `bacf869`.

---

cprussin:
## Round 3 review — recommending approval

(Posted as a plain comment: the review API is rate-limited for this account right now.)

Reviewed at head `4599dc9`. Both round-2 blockers are gone (the wlr/systemd claim in `sway/default.nix`, and the `$out/lib/systemd/user` claim plus the `systemd.packages` line in `xdg-portal/default.nix`). I re-verified the surviving claims against the pinned inputs rather than taking the fix on trust.

### The two remaining comments are accurate

**`config/modules/ui/xdg-portal/default.nix:10-12`** — both halves check out at nixpkgs `a9e6d84`:

- `/share/dbus-1`: `nixos/modules/services/system/dbus.nix:123-126` adds `[ "/etc/dbus-1" "/share/dbus-1" ]` to `environment.pathsToLink`, inside `mkIf cfg.enable`. `config/modules/system/dbus/default.nix:2-3` sets `services.dbus.enable = true`, and `config/profiles/pc/default.nix:16,44` imports both that module and this one — so it genuinely applies here.
- `/share/systemd`: present in the default `environment.pathsToLink` at `nixos/modules/config/system-path.nix:203` (the list spans 188-205, as the PR body says).

**`config/modules/ui/xdg-portal/default.nix:8-9`** ("the per-user profile is filtered by this list") — `nixos/modules/config/users-groups.nix:990-996` builds `/etc/profiles/per-user/${name}` with `inherit (config.environment) pathsToLink extraOutputsToInstall`. Correct.

### Will gtk actually start without `systemd.packages`? Yes — traced end to end

1. **The unit is built and lands in `share/`.** Upstream `meson.build` (tag 1.15.3) sets `systemd_userunit_dir = prefix / 'lib' / 'systemd' / 'user'` with **no** systemd dependency check, and `data/meson.build` installs `xdg-desktop-portal-gtk.service` there unconditionally. nixpkgs' `pkgs/by-name/xd/xdg-desktop-portal-gtk/package.nix` sets no `dontMoveSystemdUserUnits` and no overriding meson flag, and `move-systemd-user-units.sh` is a **default** stdenv fixup hook (`pkgs/stdenv/generic/default.nix:81`), so it relocates to `$out/share/systemd/user/`.
2. **It reaches the per-user profile** via `/share/systemd` (default `pathsToLink`) + the `buildEnv` above → `/etc/profiles/per-user/cprussin/share/systemd/user/xdg-desktop-portal-gtk.service`.
3. **`systemd --user` can see it.** `environment.profileRelativeSessionVariables.XDG_DATA_DIRS = [ "/share" ]` (`nixos/modules/programs/environment.nix:51`) is expanded over `environment.profiles`, which includes `/etc/profiles/per-user/$USER` (`users-groups.nix:999-1005`), and written to `/etc/pam/environment` (`nixos/modules/config/system-environment.nix`). `user@.service` runs `PAMName=systemd-user`; NixOS supplies that stack (`nixos/modules/system/boot/systemd/user.nix`, `security.pam.services.systemd-user`) and its session includes `pam_env.so conffile=/etc/pam/environment` (`nixos/modules/security/pam.nix:1479-1488`). `$XDG_DATA_DIRS/systemd/user` is in systemd's user-mode unit load path. This is the same mechanism NixOS itself banks on in the two comments in `systemd/user.nix` about `/etc/profiles/per-user/$USER/etc/xdg` being in systemd's `$XDG_CONFIG_DIRS`.
4. **The frontend argument is load-bearing, not a coincidence.** I checked the alternatives explicitly: `xdg-desktop-portal.service.in` (1.20.4, the pinned version) has **no `[Install]` section** — so no package-provided `.wants` link is even possible — no `.socket` unit, and its D-Bus service file carries `SystemdService=xdg-desktop-portal.service`. Nothing in this repo sets NixOS-level `xdg.portal` (only `primary-user.home-manager.xdg.portal`), so `nixos/modules/config/xdg/portal.nix` and its `systemd.packages` never fire. The frontend is reachable *only* through the path in (3), and it runs today. That is real evidence, not a coincidence.

`org.freedesktop.impl.portal.desktop.gtk.service.in` does name `SystemdService=xdg-desktop-portal-gtk.service` unconditionally, and the unit's `PartOf=`/`After=graphical-session.target` don't gate D-Bus activation. So the removed `systemd.packages` line was genuinely redundant.

### Other checks

- **`extraPortals` merges non-empty**: `config/modules/ui/xdg-portal/default.nix:36` contributes `[gtk]`, `config/modules/ui/sway/default.nix:213` contributes `[wlr]`, both under `primary-user.home-manager`, so HM's `assertion = cfg.extraPortals != [ ]` (`modules/misc/xdg-portal.nix:132-136` at `65258d5`) holds. HM's other assertion needs `/share/applications` and `/share/xdg-desktop-portal` in `pathsToLink` — both set at `xdg-portal/default.nix:13-16`. Empirically confirmed: `eval lyra` is green, and lyra → `profiles/laptop` → `profiles/pc` → both modules.
- **`configPackages` removal is a true no-op**: HM only feeds it into `home.packages` (`packages ++ cfg.configPackages`), which `mkForce` discards anyway; and `cfg.config` is non-empty so the "you should set `config` or `configPackages`" warning stays silent. `pkgs.sway` is in the sway `mkForce` list regardless, so its own `share/xdg-desktop-portal` entry still reaches the profile. Comment at `sway/default.nix:209-211` is accurate.
- **`config.sway.default = ["wlr" "gtk"]`** (`sway/default.nix:214`) matches the actual `.portal` basenames — `wlr.portal` and `gtk.portal`.
- **Style**: `lib` and `pkgs` are both in scope and used in each file (`lib.mkForce`, `pkgs.*`), no unused args for deadnix, nothing statix-flaggable, formatting is alejandra-shaped. `flake check` was still in progress when I looked — worth confirming green before merge, but nothing in the diff should trouble it.
- **Commit message and PR body**: re-read both against the sources; the mechanism they describe matches what I traced, including the "Corrected from earlier revisions" section.

### One optional wording nit (non-blocking)

`config/modules/ui/xdg-portal/default.nix:30` — and the matching sentence in the PR body — says gtk "implements OpenURI". There is no `org.freedesktop.impl.portal.OpenURI`; `gtk.portal` lists `FileChooser`, `AppChooser`, `Print`, `Notification`, `Inhibit`, `Access`, `Account`, `Email`, `DynamicLauncher`, `Lockdown`, `Settings`, `Wallpaper`. OpenURI is implemented in the *frontend*, which falls back to the **AppChooser** impl when there is no default handler. The conclusion is unaffected; s/OpenURI/AppChooser/ (or "AppChooser, which OpenURI falls back to") would make it exactly right. Not worth a CI round on its own — fold it in if you touch the branch again.

Nothing substantive remains. Recommend approval and merge once `flake check` goes green.

---

PR #23, opened 2026-08-29
launcher: reuse the running Claude instead of starting another copy

cprussin:
Claude Desktop closes to the tray rather than quitting, and launching it again starts a second copy instead of coming back to the one already there. The launcher's `claude` entry now points at a wrapper that raises what's running.

A closed window is gone from sway's tree entirely — the surface is destroyed, not hidden — so there are two cases:

- **Window still open** → `swaymsg [con_id=…] focus`. Matched by the owning process's argv[0] basename (`/proc/<pid>/cmdline`) rather than `app_id`, which the bundled Electron build can change between releases, and walked in sway's MRU order (descending each container's `focus`, with a plain scan appended so scratchpad windows stay reachable).
- **Only the tray icon left** → find the StatusNotifierItem whose bus connection belongs to a `claude-desktop` process, read its `Menu`, and click the entry that reopens the window — the same path a mouse click through waybar takes. `Activate`, the spec's primary action, is tried only afterwards: Electron puts the icon up through libayatana-appindicator, which serves the menu and answers `Activate` with an error.

Arguments are a deep link only a launch can hand over, so they keep the old path. A copy still starting up has neither window nor tray icon yet — what a double-tapped launcher key looks like from here — so that case polls for the window and only then falls back to `notify-send` and a launch, rather than leaving the launcher looking broken.

### Picking the menu entry

The one part that depends on a label we don't control. A wrong click is worse than none, since the D-Bus call succeeds when the app accepts the event rather than when a window appears — so it drops anything definitely something else first (quit, hide, close, settings, about, updates, devtools), then takes the best of what's left, most specific first: a plain "show"/"restore"/"reopen"; an "open" naming only the app; an "open" naming Claude; any "open"; anything naming Claude. Disabled, hidden, separator and submenu-parent entries are filtered out. If nothing qualifies it declines and launches instead of guessing.

### Testing

The D-Bus path was exercised for real, not reasoned about: a fake StatusNotifierWatcher + StatusNotifierItem + `com.canonical.dbusmenu` service (shaped like libayatana-appindicator — including its `Activate` error, and honouring `propertyNames` and `recursionDepth`) served over `dbus-run-session`, with the rendered script driven against it and stub `swaymsg`/`notify-send`/`claude-desktop` binaries.

| scenario | result |
|---|---|
| tray icon only, no window | clicks the open entry, no launch |
| window open | focuses it, tray untouched |
| invoked with a deep link | launches, hand-off preserved |
| nothing running | launches |
| running, but nothing answers | polls, notifies, then launches |

Menu-entry selection, end to end against that service:

| menu | clicked |
|---|---|
| `Open Cowork`, `Show Claude`, `Quit` | Show Claude |
| `Open Cowork`, `Open Claude Desktop` | Open Claude Desktop |
| `Open Developer Tools`, `Show Claude` | Show Claude |
| `Open Recent` (submenu), `Show Claude` | Show Claude |
| `Open Settings`, `Show Window` | Show Window |
| `_Show Claude`, `Quit` | Show Claude |
| separator, `Open Claude` | Open Claude |
| `Hide Claude` / `Show/Hide Claude` + `Quit Claude` | none → launch |
| `Quit Claude` only, empty menu | none → launch |
| `Show Claude` disabled or invisible | none → launch |

Fast paths timed to confirm the start-up poll costs them nothing: window present 0.47s, tray-only 0.50s, args 0.34s, nothing running 0.03s.

### Still worth checking on the machine

Which entry the tray menu actually offers is inferred from its label, and Claude Desktop's real menu was never visible from where this was written. If a tray-only `claude` launches a second copy and notifies instead of raising, the labels are the fix — `busctl --user --json=short get-property org.kde.StatusNotifierWatcher /StatusNotifierWatcher org.kde.StatusNotifierWatcher RegisteredStatusNotifierItems`, then `GetLayout` on that item's `Menu` path.

### DIFF
--- a/config/modules/ui/launcher/apps/claude.nix
+++ b/config/modules/ui/launcher/apps/claude.nix
@@ -1,0 +1,201 @@
+{
+  writeShellScript,
+  claude-desktop,
+  coreutils,
+  jq,
+  notify-send,
+  sway,
+  systemd,
+}:
+# Claude Desktop closes to the tray rather than quitting, and launching it again
+# starts a second copy instead of coming back to the one that's already there.
+# So don't launch it when it's already running: raise what's running instead.
+#
+# Two ways to raise it, because a closed window is gone from sway's tree
+# entirely -- it isn't hidden or minimized, the surface is destroyed -- and
+# there is nothing left to focus.  With a window still open we focus it; with
+# only the tray icon left we click that icon's own "open" entry over D-Bus,
+# which is the same path a mouse click through waybar takes.
+writeShellScript "claude" ''
+  busctl=${systemd}/bin/busctl
+  claudeDesktop=${claude-desktop}/bin/claude-desktop
+  date=${coreutils}/bin/date
+  jq=${jq}/bin/jq
+  notifySend=${notify-send}/bin/notify-send
+  sleep=${coreutils}/bin/sleep
+  swaymsg=${sway}/bin/swaymsg
+  test=${coreutils}/bin/test
+
+  # Every container lists its children in `focus` most-recently-used first, so
+  # descending that rather than `nodes` walks the windows in MRU order and the
+  # first match is the Claude window last looked at.  Nothing in the scratchpad
+  # is reachable that way -- sway builds `focus` from the seat's focus stack,
+  # which drops containers with no workspace -- so the plain scan follows as a
+  # fallback.  Repeats past that point cost nothing: the loop stops at the
+  # first Claude window either list turns up.
+  windows='
+    def leaves:
+      (.focus // []) as $order
+      | ((.nodes // []) + (.floating_nodes // [])) as $children
+      | if ($children | length) == 0
+        then .
+        else ($order[] as $id | $children[] | select(.id == $id) | leaves)
+        end;
+    (leaves | select(.pid)), (.. | objects | select(.pid and .id))
+    | "\(.pid):\(.id)"
+  '
+
+  # Pick the menu entry that reopens the window.  The label is the app's to
+  # choose and we don't get to see it from here, so drop everything that is
+  # definitely something else first -- quitting and hiding most of all -- and
+  # only then take the best of what's left, most specific first: a plain
+  # "show", then an "open" that names only the app, then one naming it at all,
+  # then any "open", then anything naming Claude.  Without the tiers a later
+  # release adding "Open Cowork" above "Show Claude" would win on menu order
+  # alone.
+  #
+  # A wrong guess is worse than no guess, because clicking is all we can see:
+  # the call succeeds when the app accepts the event, not when a window
+  # appears, so a mis-click leaves nothing raised and nothing launched either.
+  openEntry='
+    def entries:
+      .. | objects | select(.type == "(ia{sv}av)") | .data
+      | select(((.[1].visible.data) != false) and ((.[1].enabled.data) != false))
+      | select(((.[1].type.data) // "standard") != "separator")
+      | select(((.[1]["children-display"].data) // "") != "submenu")
+      | {id: .[0], label: ((.[1].label.data // "") | gsub("[_&]"; "") | ascii_downcase)};
+    [entries]
+    | map(select(.label | test("quit|exit|hide|close|sign ?out|log ?out|setting|preference|about|update|check for|developer|dev ?tools|console") | not))
+    | (map(select(.label | test("show|restore|reopen")))
+      + map(select(.label | test("^open( the)?( main)?( claude)?( app| window)?$")))
+      + map(select((.label | test("open")) and (.label | test("claude"))))
+      + map(select(.label | test("open")))
+      + map(select(.label | test("claude"))))
+    | first.id // empty
+  '
+
+  # The window belongs to the process that owns the surface, and that is the
+  # one to recognize here -- not `app_id`, which is whatever the bundled
+  # Electron build sets and can change with a release.  Match argv[0]'s
+  # basename rather than the store path, so an instance still running from
+  # before a switch is recognized too.  (Redirect stderr before opening the
+  # file: the other order reports a vanished pid on the still-open stderr.)
+  isClaude() {
+    local argv0
+    read -r -d "" argv0 2>/dev/null < "/proc/$1/cmdline" || return 1
+    case ''${argv0##*/} in
+      claude-desktop) return 0 ;;
+      *) return 1 ;;
+    esac
+  }
+
+  isRunning() {
+    local cmdline pid
+    for cmdline in /proc/[0-9]*/cmdline
+    do
+      pid="''${cmdline#/proc/}"
+      isClaude "''${pid%/cmdline}" && return 0
+    done
+    return 1
+  }
+
+  # No guard on $SWAYSOCK: with no sway reachable this yields no windows,
+  # which is the launch path anyway.
+  focusWindow() {
+    local window
+    for window in $($swaymsg -t get_tree 2>/dev/null | $jq -r "$windows")
+    do
+      if isClaude "''${window%%:*}"
+      then
+        $swaymsg "[con_id=''${window##*:}] focus" >/dev/null && return 0
+      fi
+    done
+    return 1
+  }
+
+  # Walk the tray items waybar knows about, find the one owned by Claude, and
+  # click the entry that reopens the window.  The spec's own primary action,
+  # `Activate`, would be the obvious thing to call instead, but Electron puts
+  # the icon up through libayatana-appindicator, which serves the menu and
+  # answers `Activate` with an error -- so it's only worth a try afterwards.
+  openFromTray() {
+    local item service path pid menu entry
+    for item in $($busctl --user --json=short get-property \
+      org.kde.StatusNotifierWatcher /StatusNotifierWatcher \
+      org.kde.StatusNotifierWatcher RegisteredStatusNotifierItems 2>/dev/null \
+      | $jq -r '.data[]?')
+    do
+      service="''${item%%/*}"
+      case $item in
+        */*) path="/''${item#*/}" ;;
+        *) path=/StatusNotifierItem ;;
+      esac
+
+      pid=$($busctl --user --json=short call \
+        org.freedesktop.DBus /org/freedesktop/DBus org.freedesktop.DBus \
+        GetConnectionUnixProcessID s "$service" 2>/dev/null \
+        | $jq -r '.data[0]? // empty')
+      $test "$pid" || continue
+      isClaude "$pid" || continue
+
+      menu=$($busctl --user --json=short get-property "$service" "$path" \
+        org.kde.StatusNotifierItem Menu 2>/dev/null | $jq -r '.data? // empty')
+
+      if $test "$menu"
+      then
+        # Menus are commonly filled in only when they're about to be shown.
+        $busctl --user call "$service" "$menu" \
+          com.canonical.dbusmenu AboutToShow i 0 >/dev/null 2>&1
+
+        # Depth 1, so only the entries a click can reach: a submenu's children
+        # may not be filled in until its own AboutToShow, which we never send.
+        # No property names, which the spec reads as all of them -- asking for
+        # `label` alone would leave the filters above with nothing to test.
+        entry=$($busctl --user --json=short call "$service" "$menu" \
+          com.canonical.dbusmenu GetLayout iias -- 0 1 0 2>/dev/null \
+          | $jq -r "$openEntry")
+
+        if $test "$entry" && $busctl --user call "$service" "$menu" \
+          com.canonical.dbusmenu Event isvu -- \
+          "$entry" clicked s "" "$($date +%s)" >/dev/null 2>&1
+        then
+          return 0
+        fi
+      fi
+
+      $busctl --user call "$service" "$path" \
+        org.kde.StatusNotifierItem Activate ii 0 0 >/dev/null 2>&1 && return 0
+    done
+    return 1
+  }
+
+  # Arguments are a deep link for the running instance to open, which only
+  # launching can hand over, so a bare invocation is the only one answered
+  # with a raise.
+  if $test $# -eq 0
+  then
+    focusWindow && exit
+    openFromTray && exit
+
+    # A copy still starting up has no window and no tray icon yet, which is
+    # what a double-tapped launcher key looks like from here.  Wait for it
+    # rather than answering with the second copy this exists to avoid --
+    # polling the window, since a copy that is starting maps one rather than
+    # stopping at the tray, and leaving as soon as it does.
+    if isRunning
+    then
+      for _ in 1 2 3 4 5 6 7 8
+      do
+        $sleep 0.5
+        focusWindow && exit
+      done
+      openFromTray && exit
+
+      # Launching is still better than doing nothing, but say so.
+      $notifySend -i claude-desktop "Claude is already running" \
+        "Couldn't reach its window or tray icon -- starting another copy."
+    fi
+  fi
+
+  exec $claudeDesktop "$@"
+''
--- a/config/modules/ui/launcher/default.nix
+++ b/config/modules/ui/launcher/default.nix
@@ -57,7 +57,7 @@ in {
           chatgpt = "${pkgs.chatgpt-desktop}/bin/chatgpt";
           chrome = pkgs.writeShellScript "chrome" "${pkgs.launcher}/bin/browse --browser chrome $*";
           chromium = pkgs.writeShellScript "chromium" "${pkgs.launcher}/bin/browse --browser chromium $*";
-          claude = "${pkgs.claude-desktop}/bin/claude-desktop";
+          claude = pkgs.callPackage ./apps/claude.nix {};
           credit-cards = mkWebApp "credit-cards" "https://docs.google.com/spreadsheets/d/1Y8xind-5nMe9bezMFmk__CQdkSBd7FPupt1NkdKDLUE?authuser=connor@prussin.net";
           crux = mkTerminalApp "crux" "${pkgs.openssh}/bin/ssh -t crux load-session";
           cups = mkWebApp "cups" "http://localhost:631";


config/modules/ui/launcher/apps/claude.nix
```diff
@@ -0,0 +1,46 @@
...
+# Claude Desktop closes to the tray rather than quitting, and it holds an
+# Electron single-instance lock, so re-running the binary while it is up hands
+# off to the instance that's already there instead of drawing anything.  When
+# that instance still has a window somewhere -- another workspace, or just
+# behind whatever is focused -- focus it ourselves and skip the hand-off.
+#
+# Windows are matched by the pid that owns them rather than by `app_id`: the
+# app_id is whatever the bundled Electron build happens to set and a release can
+# change it under us, while the surface always belongs to the claude-desktop
+# process.
```
cprussin: The premise here is load-bearing and unverified, and the two ways it can resolve point in opposite directions:

- If the second-instance hand-off genuinely draws nothing, then the case you actually hit — window closed to the tray — is **still broken after this PR**. A hidden window is not a mapped surface, so it is not in `get_tree`; `conId` stays empty and we fall through to `exec $claudeDesktop`, i.e. exactly the no-op being fixed. This change only covers "a window exists but isn't focused".
- If the hand-off *does* raise the window, then the only real bug was that it doesn't switch workspaces, and the whole wrapper is a smaller fix than it looks.

Worth 30 seconds on lyra before merging: (a) open Claude, move it to another workspace, run `claude` — does the stock binary raise it? (b) close it to the tray, run `claude` — does anything appear? The answers decide whether this is done or whether the fallback needs the SNI `Activate` poke you mention in the description.

Either way I'd amend the comment to say what the wrapper covers and what it doesn't, since the tray sentence currently reads as though the tray case is handled.

---

config/modules/ui/launcher/apps/claude.nix
```diff
@@ -0,0 +1,46 @@
...
+  if $test "$SWAYSOCK"
+  then
+    for window in $($swaymsg -t get_tree | $jq -r '.. | objects | select(.pid and .id) | "\(.pid):\(.id)"')
+    do
+      if $grep -qsz claude-desktop "/proc/''${window%%:*}/cmdline"
+      then
+        conId="''${window##*:}"
+        break
+      fi
+    done
```
cprussin: `break` takes the first match in *tree order*, which is an arbitrary choice once there is more than one Claude window — and with Cowork that is the normal state, not an edge case. Tree order is output → workspace → container nesting, and the scratchpad workspace is in that tree as well, so "first" is roughly "whichever window happens to sit earliest in the layout", never "the one I was last looking at". Repeated `claude` invocations will keep yanking you to the same window even if you just came from another one.

Two cheap improvements, either is fine:

- Prefer the most-recently-focused: sway's `focus` array at each level is ordered MRU, so walking `.focus[0]` down from root gives you the MRU leaf, and filtering the pid-matched set by that order gets you "the Claude window I used last".
- Or at minimum prefer a match on the currently-visible workspace before falling back to the first one found.

Also note the loop shells out to `grep` once per window on the machine; collecting the candidate pids first (`pgrep -x claude-desktop`, which fits in `comm`'s 15 chars) and filtering the tree with a single `jq --argjson` pass would be both cheaper and easier to read, if you'd rather restructure than special-case.

---

config/modules/ui/launcher/apps/claude.nix
```diff
@@ -0,0 +1,46 @@
...
+  swaymsg=${sway}/bin/swaymsg
+  test=${coreutils}/bin/test
+
+  conId=""
+
+  if $test "$SWAYSOCK"
+  then
+    for window in $($swaymsg -t get_tree | $jq -r '.. | objects | select(.pid and .id) | "\(.pid):\(.id)"')
+    do
+      if $grep -qsz claude-desktop "/proc/''${window%%:*}/cmdline"
```
cprussin: The match is an unanchored substring against every argv element, so any window whose owning client's command line merely *mentions* `claude-desktop` gets focused instead — e.g. a terminal started as `foot -e claude-desktop ...`, or anything spawned with a path under `.../claude-desktop/...`. Narrow, but the fix is free: `$grep -qszx` won't work (argv[0] is a full path), but anchoring on the store path does — `$grep -qsz -F "${claude-desktop}"`, since the real process argv[0] is `<store>/lib/claude-desktop/claude-desktop` and the wrapper's is `<store>/bin/claude-desktop`.

Trade-off worth weighing before you take that: after a `switch` that bumps the package, the *running* app is still the old store path, so a store-path match would stop recognizing it and you'd relaunch (back to the no-op). The bare-name match survives upgrades. If you keep the bare name, that reasoning belongs in the comment next to the `app_id` rationale, since it's the same class of decision.

`-qsz` semantics themselves are right: `-z` splits on NUL so each argv element is a record, and `-s` swallows the read error when the pid exits between `get_tree` and the `grep`.

---

config/modules/ui/launcher/apps/claude.nix
```diff
@@ -0,0 +1,46 @@
...
+writeShellScript "claude" ''
+  claudeDesktop=${claude-desktop}/bin/claude-desktop
+  grep=${gnugrep}/bin/grep
+  jq=${jq}/bin/jq
+  swaymsg=${sway}/bin/swaymsg
+  test=${coreutils}/bin/test
+
+  conId=""
+
+  if $test "$SWAYSOCK"
```
cprussin: Gating on `$SWAYSOCK` is stricter than `swaymsg` itself, which also honours `$I3SOCK` and falls back to `sway --get-socketpath`. Any launch path that doesn't inherit `SWAYSOCK` (a systemd user unit, a `.desktop` activation from a portal) silently skips the focus attempt and relaunches. Since the failure mode is already "just run `swaymsg` and see", you could drop the guard and let the pipeline produce nothing when there's no sway — the loop over an empty `$(...)` is already the no-window path. If you keep the guard, `swaymsg -t get_tree >/dev/null 2>&1` as the condition is a truer test than the env var.

---

config/modules/ui/launcher/apps/claude.nix
```diff
@@ -0,0 +1,46 @@
...
+      then
+        conId="''${window##*:}"
+        break
+      fi
+    done
+  fi
+
+  if $test "$conId"
+  then
+    exec $swaymsg "[con_id=$conId] focus"
```
cprussin: `"$@"` is silently dropped on this branch. The launcher does pass arguments through — `pkgs/launcher/scripts/run.nix` execs `$HOME/.launcher-apps/$@`, so typing `claude <something>` in the launcher reaches this script with args. Before this change those args went to `claude-desktop` (and thence to the running instance's second-instance handler, which is how Electron apps take deep links); now they vanish and you just get a focus.

Cheapest correct thing: only take the focus path when there's nothing to hand off — `if $test "$conId" -a $# -eq 0`. Or focus *and* hand off. Low stakes since `claude` is normally invoked bare, but the silent drop is the kind of thing that's confusing when it does bite.

---

cprussin (commented):
Mechanically this is clean and idiomatic for the repo — the `''${window%%:*}` escapes are right, `\(.pid)` passes through the `''` string untouched (same as `screen.nix`), the callPackage argument set matches exactly what's used, `claude-desktop` resolves because the overlay is applied in the same module's `nixpkgs.overlays` (and `eval lyra` is green), and the `$test`/`$grep`-style bound-path prelude matches `screen.nix`. `writeShellScript` gives you bash, so `''${var%%:*}` is fine (it's POSIX anyway). `swaymsg "[con_id=N] focus"` is the right syntax and does switch workspaces.

My concerns are about what it achieves, not whether it runs:

1. **The stated case probably isn't fixed.** A window hidden to the tray is an unmapped surface and doesn't appear in `get_tree`, so that path falls straight through to `exec $claudeDesktop` — the behaviour being replaced. This wrapper covers "window exists but isn't focused", nothing more. Please confirm on lyra which of the two states actually misbehaves before merging; that also tells you whether the wrapper is needed at all.
2. **Multiple windows are resolved arbitrarily.** `break` on the first tree-order match is not "the Claude window I was last using", and with Cowork more than one window is routine.
3. Minor: the pid substring match is broader than intended, the `$SWAYSOCK` guard is stricter than `swaymsg` needs, and `"$@"` is silently dropped on the focus branch.

Nothing here is a blocker on its own, but (1) decides whether the PR is finished.

---

cprussin:
Force-pushed `2b517a1` addressing 2–5:

- **MRU selection** — the jq pass now descends each container's `focus` array instead of `nodes`, so windows come out most-recently-used first and the first match is the Claude window last looked at. Verified against a synthetic tree with two outputs, a floating window and an empty output.
- **Unanchored pid match** — now `head -z -n 1 /proc/<pid>/cmdline | grep -Eqz '(^|/)claude-desktop$'`, i.e. argv[0]'s basename only. `foot -e …/claude-desktop` no longer matches. Kept the basename rather than the store path for exactly the reason raised — an instance running from an older generation after a `switch` must still be recognized — and that trade-off is now in the header comment.
- **`$SWAYSOCK` guard** — dropped. `swaymsg -t get_tree 2>/dev/null` producing nothing is already the launch path.
- **`"$@"` dropped** — focus branch is gated on `$# -eq 0`; anything with arguments takes the hand-off, which is the only path that can deliver a deep link.

On **1**, the load-bearing one: you're right that a tray-hidden window isn't in the tree and falls through unchanged, and I can't settle it from here — `downloads.claude.ai` is blocked by this environment's egress policy, so `app.asar` isn't inspectable. The header comment now says plainly what the wrapper covers and what it doesn't, rather than implying the tray case is handled. The two-state test on lyra is the next step before merging; if (b) shows nothing, the fallback needs the SNI `Activate` / dbusmenu poke and that's a separate commit.

---

config/modules/ui/launcher/apps/claude.nix
```diff
@@ -0,0 +1,70 @@
...
+  mruWindows='
+    def leaves:
+      (.focus // []) as $order
+      | ((.nodes // []) + (.floating_nodes // [])) as $children
+      | if ($children | length) == 0
+        then .
+        else ($order[] as $id | $children[] | select(.id == $id) | leaves)
+        end;
+    leaves | select(.pid) | "\(.pid):\(.id)"
+  '
```
cprussin: The MRU traversal is correct for everything reachable, but it loses one case the old `..` scan covered: **a Claude window in the scratchpad**.

sway synthesizes the scratchpad as an output named `__i3` with id `INT32_MAX` (`sway/ipc-json.c`, `i3_output_id`), and `describe_node_recursive` puts it first in root's `nodes`. But root's `focus` array is built by walking the seat's focus stack and mapping each entry through `node_get_output()` — and a scratchpad-hidden container has `pending.workspace == NULL`, so `node_get_output()` returns NULL and it's dropped. The synthetic `__i3` id is not a real node and is never in the focus stack either. Net: `__i3` is in `nodes` but never in `focus`, so `leaves` never descends into it.

That matters because `[con_id=N] focus` *does* handle hidden scratchpad containers — `cmd_focus` calls `root_scratchpad_show(container)` before `seat_set_focus_container`. So before this push, a Claude window sent to the scratchpad got un-hidden and focused; now it falls through to the hand-off, which is the no-op this PR exists to fix.

I checked this against a synthetic tree shaped like the real one:

```
tree                      | old `..`      | new `leaves`
--------------------------|---------------|-------------
scratchpad-hidden claude  | found         | missed
tiled claude (MRU 1st)    | found (2nd)   | found (1st)  ✓
floating claude           | found         | found        ✓
tabbed children           | tree order    | MRU order    ✓
```

Everything else I tried to break it with holds up: a top-level floating container's `node_get_parent()` is its workspace, so it *is* in the workspace's `focus` (I'd assumed otherwise until I read `sway/tree/node.c`); every node except `N_ROOT`/`N_OUTPUT` gets a seat node at creation, so nothing is missing from `focus` merely for never having been focused (it just sorts last); root dedupes outputs; a `$order` id with no matching child yields nothing rather than erroring. So the scratchpad is the only hole.

Narrow enough that "I don't scratchpad Claude" is a fine answer — but if you'd rather not lose it, appending the unordered scan as a fallback keeps MRU preference and costs one line, since duplicates just cost an extra `grep` and the loop breaks on the first hit either way:

```
leaves | select(.pid) | "\(.pid):\(.id)"
```
becomes
```
(leaves | select(.pid)), (.. | objects | select(.pid and .id))
| "\(.pid):\(.id)"
```

---

config/modules/ui/launcher/apps/claude.nix
```diff
@@ -0,0 +1,70 @@
...
+
+  conId=""
+
+  # No guard on $SWAYSOCK: swaymsg finds sway by other means too, and with no
+  # sway at all this just yields no windows, which is the launch path anyway.
```
cprussin: Dropping the guard is the right call, but "swaymsg finds sway by other means too" is not quite what happens, and my earlier comment was the source of the error — worth correcting so the next reader doesn't rely on it.

`get_socketpath()` in `common/ipc-client.c` tries, in order: `$SWAYSOCK`, then `popen("sway --get-socketpath")`, then `$I3SOCK`, then `popen("i3 --get-socketpath")`. And `sway --get-socketpath` is just:

```c
case 'p': // --get-socketpath
    if (getenv("SWAYSOCK")) {
        printf("%s\n", getenv("SWAYSOCK"));
        exit(EXIT_SUCCESS);
    } else {
        fprintf(stderr, "sway socket not detected.\n");
        exit(EXIT_FAILURE);
    }
```

It re-reads `SWAYSOCK` from its own (inherited) environment, so it finds nothing that the first check didn't. The only genuinely additional path is `$I3SOCK`, which sway itself never sets. So the systemd-unit / portal-activation case I cited still doesn't work — an environment without `SWAYSOCK` still gets no tree and still relaunches.

The change is still a net positive (one less thing to keep in sync, and the empty-tree path was already the launch path), and the cost when there's no sway is two `popen`s of a program that exits immediately. Just reword to something like "no guard on `$SWAYSOCK`: with no sway reachable this yields no windows, which is the launch path anyway" and drop the claim about other means.

---

cprussin (commented):
Re-reviewed at `2b517a1`. Four of the five earlier findings are addressed properly; the fifth (verify on lyra what the stock hand-off actually does) is now at least honestly documented in the header comment rather than papered over. Two things left, both minor — nothing blocking.

**What I verified, and it holds:**

*Nix.* `''${window%%:*}` / `''${window##*:}` render to `${...}` correctly; `\(.pid)` is a literal backslash-paren in an indented string (only `''\` escapes there) and reaches jq inside single quotes untouched; the multi-line `mruWindows='...'` contains no `'` and no `''`, so neither layer bites. `callPackage` args exactly match usage — `head` and `test` from `coreutils`, `grep` from `gnugrep`, plus `jq`, `sway`, `claude-desktop`, `writeShellScript`; nothing unused. I rendered the string by hand and ran `bash -n` on it: clean, which is also what `writeShellScript`'s `shellDryRun` checkPhase will do at build time. Formatting matches alejandra's output and the sibling `apps/*.nix` files.

*The pid match.* `head -z -n 1` emits exactly `argv[0]\0`, and `grep -Eqz '(^|/)claude-desktop$'` anchors `$` at the record's NUL. Tested against real `/proc` entries created with `exec -a`:

| argv[0] | result |
|---|---|
| `/nix/store/…/bin/claude-desktop` (wrapper) | match |
| `/nix/store/…/lib/claude-desktop/claude-desktop` (real binary) | match |
| `/usr/bin/foot` with `claude-desktop` in argv[2] | no match ✓ (this was the old bug) |
| `/opt/not-claude-desktop` | no match ✓ |
| empty cmdline | no match |
| nonexistent pid | no match, exit 1 |

Dropping `-s` is fine — `grep` reads a pipe, and the `2>/dev/null` is on the `head` that actually opens the file; pipeline status is `grep`'s either way. Helper processes can't shadow the main one: sway takes a view's `pid` from the Wayland client credentials, and under Ozone/Wayland only the browser process creates toplevels, so a renderer or GPU pid never appears in the tree in the first place (and would match anyway, harmlessly). Keeping the basename rather than the store path is the right trade for the post-`switch` case, and the header now says so.

*Argument gating.* `$test "$conId" && $test $# -eq 0` is correct — `$#` is always a bare integer so the missing quotes are inert. And it really is 0 args in the common case: `run.nix` does `exec $HOME/.launcher-apps/$@` with `$@` word-split, so a bare `claude` reaches this script with nothing, while `claude <deep-link>` reaches it with one arg and takes the hand-off. No `.desktop` entry or keybinding invokes it with `%u`, so there's no other caller to worry about.

*End to end.* Driving the rendered script with a fake `swaymsg` and real pids: bare invocation focuses the MRU Claude window, flipping the workspace's `focus` order flips which one it picks, args always hand off, and no sway at all exits 0 down the launch path.

**Nothing in the four fixes made the common case worse.** The MRU walk is pure jq; the anchored match is strictly narrower; the `$#` gate only diverts a path that was previously silently broken; and the dropped `SWAYSOCK` guard costs nothing when sway is running (the first `getenv` hits).

Two comments inline: a narrow scratchpad regression the MRU traversal introduces versus the old `..` scan, and a correction to the `$SWAYSOCK` rationale in the comment — which was my error in the last round, not yours.

---

cprussin:
Force-pushed `a79a078`. All five taken.

**1. `GetLayout` asked for `label` only, so the filters were inert.** Correct, and the harness was green because the fake ignored `propertyNames` — my error, not the jq's. The fake now honours both `propertyNames` and `recursionDepth` like a compliant server, and with that a disabled or hidden `Show Claude` is correctly skipped where it previously would have been clicked. Request is now `GetLayout iias -- 0 1 0`: empty list for all properties, and depth 1 per your other point, which also kills the stale-submenu-child problem outright.

**2. Submenu parents.** Depth 1 keeps children out of the reply, and `children-display != "submenu"` (now visible, since properties are actually requested) drops the parent itself, so `Open Recent` no longer wins.

**3. Entry picker.** Adopted the tiered version, plus `hide|close|developer|dev ?tools|console` on the exclusion list. Every row of your table reproduces end-to-end against the fake service:

| menu | clicked |
|---|---|
| `Open Cowork`, `Show Claude`, `Quit` | Show Claude |
| `Open Developer Tools`, `Show Claude` | Show Claude |
| `Hide Claude`, `Quit Claude` | none → launch |
| `Open Recent` (submenu), `Show Claude` | Show Claude |
| `_Show Claude`, `Quit` | Show Claude |
| `Open Claude`, `Quit` | Open Claude |
| `Open Settings`, `Show Window` | Show Window |
| `Quit Claude` only | none → launch |
| empty menu | none → launch |
| `Show Claude` disabled | none → launch |
| `Show Claude` invisible | none → launch |
| separator, `Open Claude` | Open Claude |

**4. Start-up race.** Real, and it produces the duplicate in the likeliest way to reach it. `isRunning` now waits and retries both raise paths before the notification, and only then launches. Kept launching over refusing, for the reason you gave.

**5. `isRunning` forks and timestamp `0`.** Took the fork-free `read -r -d ""` version (with the redirection ordered as you noted), which drops the `head`/`gnugrep` bindings and the `gnugrep` callPackage argument entirely. Timestamp is now `date +%s`.

The four original scenarios still hold on the new head: tray-only clicks and doesn't launch, an open window is focused with the tray untouched, args hand off, nothing-running launches.

Nothing to do about the earlier threads you couldn't resolve — the rate limit was on your side, and they're obsolete either way.

---

cprussin:
Took both optional ones — `1a6921e`.

**`Open Claude Desktop` losing to `Open Cowork`.** Added the `open`-and-`claude` tier between the anchored one and the loose one. Both new rows behave, and the whole table is unmoved:

| menu | clicked |
|---|---|
| `Open Cowork`, `Open Claude Desktop` | Open Claude Desktop |
| `Open Cowork`, `Open Claude` | Open Claude |
| `Show/Hide Claude`, `Quit Claude` | none → launch |

The combined toggle declining to guess is the behaviour I want, for the reason you gave — at the moment `openFromTray` runs there's no window, so a state-reflecting label can read either way.

**`sleep 2` shape.** Replaced with the bounded poll: eight rounds of `sleep 0.5` on `focusWindow`, then one `openFromTray`. Leaves as soon as the window maps instead of paying the full budget, and gives a cold Electron start 4s instead of 2. Timed the paths that shouldn't be touched — window present 0.47s, tray-only 0.50s, args 0.34s, nothing-running 0.03s.

Also noted your `read` divergence on an argv[0] with no trailing NUL. Leaving it: fails safe, and the kernel terminates `cmdline` for anything that hasn't rewritten its argv area.

---

config/modules/ui/launcher/apps/claude.nix
```diff
@@ -0,0 +1,201 @@
...
+      .. | objects | select(.type == "(ia{sv}av)") | .data
+      | select(((.[1].visible.data) != false) and ((.[1].enabled.data) != false))
+      | select(((.[1].type.data) // "standard") != "separator")
+      | select(((.[1]["children-display"].data) // "") != "submenu")
+      | {id: .[0], label: ((.[1].label.data // "") | gsub("[_&]"; "") | ascii_downcase)};
+    [entries]
+    | map(select(.label | test("quit|exit|hide|close|sign ?out|log ?out|setting|preference|about|update|check for|developer|dev ?tools|console") | not))
+    | (map(select(.label | test("show|restore|reopen")))
+      + map(select(.label | test("^open( the)?( main)?( claude)?( app| window)?$")))
+      + map(select((.label | test("open")) and (.label | test("claude"))))
```
cprussin, line 71: Tier is correct and can't regress the tier below it in the general case: the set it selects is a strict subset of the `test("open")` tier, so concatenating it first only re-ranks within that set — it can never surface an entry that tier 4 would have rejected, and the duplicates it leaves behind are harmless since only `first` is read.

I reproduced the tier chain standalone and confirmed the intended fix plus the surrounding rows:

| menu | picked |
|---|---|
| `Open Cowork` / `Open Claude Desktop` | Open Claude Desktop |
| `Open Cowork` / `Open Claude` | Open Claude |
| `Show/Hide Claude` / `Quit Claude` | none |
| `Reopen Claude` / `Open Cowork` | Reopen Claude (tier 1) |
| `Open Window` / `Open Claude Support` | Open Window (tier 2, anchored) |
| `Open Claude Cowork` / `Open Claude Desktop` | Open Claude Cowork (menu order, unavoidable) |

On the substring question: both `test`s are unanchored, but the labels are `ascii_downcase`d beforehand so case isn't a hazard, and the only realistic word containing `open` as a substring is `reopen` — which tier 1 already claims before this tier is consulted. `claude` isn't a substring of anything that isn't Claude.

One behaviour change beyond the intended fix, non-blocking: where both candidates are unanchored `open` entries and only one names the app, menu order no longer decides. e.g. `Open Main Window Now` / `Open Claude Support` now picks the latter, where before it picked the former. Both are guesses at that point, and "names the app" is the better prior than "is listed first", so I'd leave it — just noting it's a real, if unlikely, swing.

---

config/modules/ui/launcher/apps/claude.nix
```diff
@@ -0,0 +1,201 @@
...
+    # polling the window, since a copy that is starting maps one rather than
+    # stopping at the tray, and leaving as soon as it does.
+    if isRunning
+    then
+      for _ in 1 2 3 4 5 6 7 8
+      do
+        $sleep 0.5
+        focusWindow && exit
+      done
+      openFromTray && exit
```
cprussin, line 192: Loop checks out on every point raised. Verified locally on bash 5.2 / coreutils 9.4:

- **`_` as the loop variable** — safe. `_` is an ordinary writable variable that bash happens to reassign after every simple command; a `for` assigning it is idiomatic and has no side effect here, since nothing in the script reads `$_`. Even inside the loop the value doesn't survive: bash clobbers it with the last argument of each command as it runs (`for _ in 1 2; do :; done; echo "$_"` prints `:`, not `2`).
- **`$sleep 0.5`** — resolves to `${coreutils}/bin/sleep`, which takes fractional arguments; there is no bash `sleep` builtin to shadow it, so this is unambiguous either way.
- **Budget** — 8 × 0.5s = 4.0s of sleep measured, plus one `swaymsg`+`jq` per iteration, so ~4.1s worst case before falling through. The reworded comment makes no numeric claim, so nothing to keep in sync.
- **Early-exit paths** — none got slower. The first probe now lands at 0.5s instead of 2s, and the sleep-before-check ordering is right: `focusWindow` already ran immediately above the `isRunning` guard, so checking before the first sleep would just repeat it.

Leaving `openFromTray` outside the loop is the right call, not an oversight. It has side effects the window poll doesn't — it clicks a menu entry chosen by a heuristic, and a label like `Show/Hide Claude` is a toggle, so retrying it on a cadence risks showing and then re-hiding the window. Running it once after the poll still covers the case the comment doesn't cover (a copy that starts straight to the tray without ever mapping a window): by then the icon has had 4s to register, and the single click lands. Cost of that choice is only the 4s spent polling first, on what should be a rare path.

The one thing the loop doesn't re-check is `isRunning`, so a copy that crashes mid-startup burns the full 4s before launching. Same eventual outcome, just later, on an unlikely path — not worth the extra `/proc` scan per iteration.

---

cprussin (commented):
Both follow-ups landed cleanly. No blocking issues, and nothing new turned up on a whole-file re-read — this reads as ready to merge from my side. (Submitting as a comment rather than an approval only because GitHub won't let the PR author's token approve its own PR.)

**New jq tier (line 71)** — correct and safe by construction: its selection is a strict subset of the `test("open")` tier below it, so it can only re-rank within that set, never surface something that tier would have rejected. Duplicates across the two tiers are harmless because only `first` is read. Reproduced the tier chain standalone against the reported menu table and it matches, including the `Open Cowork` / `Open Claude Desktop` fix and the correct no-match on `Show/Hide Claude` / `Quit Claude`. Substring behaviour is fine — labels are `ascii_downcase`d first, and the only word realistically containing `open` is `reopen`, which tier 1 claims before this tier is reached. Details and the one non-blocking behaviour swing are in the inline comment.

**Poll loop (lines 187-191)** — verified on bash 5.2 / coreutils 9.4. `_` is a safe loop variable (bash reassigns it after every command anyway, and nothing here reads `$_`); `$sleep` is coreutils, so `0.5` works; measured budget is 4.0s of sleep plus a probe per iteration, and the comment makes no numeric claim to drift out of sync. No previously-early path got slower — the first probe moves from 2s to 0.5s, and sleeping before the first check avoids repeating the `focusWindow` that just ran above the `isRunning` guard.

**`openFromTray` staying outside the loop** — right, and worth stating as deliberate: it clicks a heuristically-chosen menu entry, and a toggle label (`Show/Hide Claude`) would flip the window back off if retried on a cadence. Polling the side-effect-free check and clicking once at the end is the correct split.

**Nix mechanics** — clean. Nothing added needs `''` escaping (no `${` or `''` in the new lines), the argument set still matches usage exactly (all seven consumed, none missing), and the diff touches only lines inside the indented string with matching relative indentation, so alejandra has nothing to move — no tabs, no trailing whitespace, new line comfortably under the file's 80-col habit. I don't have alejandra in this environment to run it directly; `cli test` covers it.

**Whole file** — read end to end. `isClaude`'s stderr-before-stdin redirect ordering, the MRU `focus` descent with the plain-scan fallback, the depth-1 `GetLayout` with no property filter, the `Activate`-only-afterwards fallback, and the notify-then-launch bailout all still read correctly and consistently with their comments. Nothing left from my side.

---

PR #24, opened 2026-08-29
[printers] Wait for wireguard before ensuring the printer

cprussin:
## Problem

`ensure-printers` fails on every boot of lyra.

- The unit runs `lpadmin -p circinus -v ipp://circinus.internal.prussin.net/... -m everywhere -E`.
- `-m everywhere` makes cupsd fetch the queue's capabilities from the server over IPP, so it needs to resolve *and* reach the host.
- `circinus.internal.prussin.net` only resolves via crux's nameserver (installed by the tunnel's `postSetup` resolvconf) and CUPS on crux only listens on the wireguard addresses behind an `@IF(prussinnet)` firewall rule.
- Upstream's unit is ordered only `after = [ "cups.service" ]`, so it races the tunnel and exits non-zero.
- Nothing persists `/var/lib/cups` across a boot, so the failure leaves the machine with no printer until the unit is started by hand.

## Fix

- Order `ensure-printers` after `wireguard-prussinnet.target`.
- Retry on failure every 15s, capped at 20 attempts. Ordering alone isn't sufficient on a laptop: the peer unit is `oneshot` and completes as soon as `wg set` returns, well before wifi has associated and the handshake has landed — so the retry is what actually makes this work.

### Why `StartLimitIntervalSec = "infinity"`

It caps *total* attempts. A finite window can't bound this: systemd resets the window whenever a start arrives after it has elapsed, and a lookup against the tunnel's black-holed nameservers takes tens of seconds, so the burst would never trip.

Bounding the attempt with `TimeoutStartSec` instead would introduce a worse bug. In CUPS 2.4.19 `-m everywhere` is not fetched by `lpadmin` — cupsd does it in a detached thread (`create_local_bg_thread`, 30s connect timeout) that outlives a SIGTERMed `lpadmin`. Every failure path there sets `state_time = 0; temporary = 1`, the success path never clears `temporary`, and `cupsdDeleteTemporaryPrinters(0)` reaps on `state_time < now - 300`. So a stale thread failing *after* a later attempt succeeded deletes the queue that just got created — no printer, green unit.

The cost is that once the cap is spent, a start by hand needs `systemctl reset-failed` first. That's noted in the comment.

Booting away from any network still ends in a failed unit, as before — no regression there.

https://claude.ai/code/session_01Dm9eWzFzp6E27QH53TJrfs

---

### DIFF
--- a/config/modules/system/devices/printers/default.nix
+++ b/config/modules/system/devices/printers/default.nix
@@ -12,4 +12,28 @@ _: {
       }
     ];
   };
+
+  # `-m everywhere` fetches the queue's capabilities over IPP, which only works
+  # once the wireguard tunnel carries traffic -- and on a laptop that's a while
+  # after the tunnel's units finish, so retry too.
+  systemd.services.ensure-printers = {
+    wants = ["wireguard-prussinnet.target"];
+    after = ["wireguard-prussinnet.target"];
+    serviceConfig = {
+      Restart = "on-failure";
+      RestartSec = "15s";
+    };
+
+    # Caps total attempts rather than attempts per window: a failure against the
+    # tunnel's black-holed nameservers takes long enough to reset a finite
+    # window instead of tripping it.  Not `TimeoutStartSec` -- `-m everywhere`
+    # is fetched by cupsd in a thread that outlives a killed lpadmin, and a
+    # stale one failing after a later attempt succeeded marks the new queue
+    # temporary, so cupsd deletes it.  Hitting the cap means a start by hand
+    # needs `systemctl reset-failed` first.
+    unitConfig = {
+      StartLimitIntervalSec = "infinity";
+      StartLimitBurst = 20;
+    };
+  };
 }


config/modules/system/devices/printers/default.nix
```diff
@@ -12,4 +12,20 @@ _: {
       }
     ];
   };
+
+  # `-m everywhere` fetches the queue's capabilities over IPP, which only works
+  # once the wireguard tunnel carries traffic -- and on a laptop that's a while
+  # after the tunnel's units finish, so retry too.
+  systemd.services.ensure-printers = {
+    wants = ["wireguard-prussinnet.target"];
+    after = ["wireguard-prussinnet.target"];
```
cprussin, line 21: Unit name checks out. In the pinned nixpkgs (`nixpkgs` input rev `d407951`; the file is byte-identical in the 26.05 input `a9e6d84`), `nixos/modules/services/networking/wireguard.nix` ends with `systemd.targets = mkIf (!cfg.useNetworkd) (mapAttrs' generateInterfaceTarget cfg.interfaces)`, and `generateInterfaceTarget` is `nameValuePair "wireguard-${name}"` — so `wireguard-prussinnet.target` is correct, and nothing in this repo sets `networking.wireguard.useNetworkd`.

The target is `wants` + `after` the interface unit and every peer unit, so the ordering does what the commit message says — but only that far. The peer unit is `Type=oneshot` + `RemainAfterExit`, so it completes as soon as `wg set … peer …` returns, well before any handshake. Its `After=network-online.target` doesn't add much here either: `wlan*` is in `networking.dhcpcd.denyInterfaces` (`config/modules/system/devices/wifi/default.nix`) and iwd isn't a wait-online provider, so nothing on this path waits for association. That confirms the premise — the retry is what actually makes this work, the ordering is cheap insurance. Worth keeping in mind if the retry is ever tuned down on the assumption the ordering gates on connectivity.

`wants` is redundant with the target's own `wantedBy = ["multi-user.target"]`, but it's harmless and it makes a manual `systemctl start ensure-printers` pull the tunnel up, so I'd keep it.

---

config/modules/system/devices/printers/default.nix
```diff
@@ -12,4 +12,20 @@ _: {
...
+    serviceConfig = {
+      Restart = "on-failure";
+      RestartSec = 15;
+    };
+    unitConfig = {
+      StartLimitIntervalSec = 600;
+      StartLimitBurst = 20;
+    };
```
cprussin, line 37: `Restart=on-failure` is legal here: systemd rejects only `always`/`on-success` for `Type=oneshot` (`src/core/service.c`: *"Service has Restart= set to either always or on-success, which isn't allowed for Type=oneshot services"*), and it composes fine with the `RemainAfterExit=true` upstream sets.

**The retry bound is the one thing I'd fix.** systemd's start ratelimit resets its window whenever a start arrives more than `StartLimitIntervalSec` after the window opened (`ratelimit_below()` in `src/basic/ratelimit.c`), so the limit only fires if all 20 starts land inside 600s:

```
19 × (D + RestartSec) ≤ 600s   →   D ≤ ~16.5s
```

where `D` is how long one `lpadmin` attempt takes to fail. That holds when the failure is fast (NXDOMAIN from a working resolver, or a prompt connect refusal), and then you get the intended ~20 attempts over ~285s. But the case this PR defends against is the slow one: once the tunnel's `postSetup` has installed crux's two wireguard addresses as nameservers and the tunnel is silently dropping packets, glibc's default `timeout:5 attempts:2` across two nameservers is on the order of 20–40s per lookup, before `lpadmin` even gets to the IPP connect. Past ~16.5s per attempt the window keeps resetting, the limit never fires, and the unit retries every ~15s indefinitely — journal noise and a unit stuck in `activating (auto-restart)` rather than the bounded failure the commit describes.

Deterministic fix: bound the attempt too, e.g. add `TimeoutStartSec = 10;` to `serviceConfig` (a start timeout counts as a failure, so `Restart=on-failure` still applies). Cycle then caps at 25s, 20 starts fit in ~475s < 600s, and the "few minutes" wording is true regardless of DNS behaviour. Alternatively raise `StartLimitIntervalSec` past 20 × worst-case cycle — but then it's several minutes longer than advertised.

Three smaller notes:

- Once the burst is spent, further starts are refused for the remainder of the window, so the documented fallback of starting it by hand needs `systemctl reset-failed ensure-printers` first if attempted within ~5 minutes of boot. Same trap the wluma comment in `config/modules/system/devices/backlight/default.nix` already spells out; a sentence here would be consistent with that.
- `RestartSec = 15` works (rendered as seconds), but sibling units in this repo write `"5s"` — a unit'd string would be more consistent.
- `unitConfig.StartLimit*` is correct (both keys belong in `[Unit]` on modern systemd), and it matches the backlight precedent; nixpkgs also exposes first-class `startLimitIntervalSec` / `startLimitBurst` options on `systemd.services.<name>`. Purely stylistic, no behavioural difference.

---

config/modules/system/devices/printers/default.nix
```diff
@@ -12,4 +12,20 @@ _: {
       }
     ];
   };
+
+  # `-m everywhere` fetches the queue's capabilities over IPP, which only works
+  # once the wireguard tunnel carries traffic -- and on a laptop that's a while
+  # after the tunnel's units finish, so retry too.
+  systemd.services.ensure-printers = {
```
cprussin, line 19: Merge semantics verified against the pinned `nixos/modules/hardware/printers.nix`: the unit is defined under `mkIf (cfg.ensurePrinters != [ ] && config.services.printing.enable)`, and both conditions are satisfied unconditionally by lines 2–14 of this same file, so the upstream definition is always present alongside this one. Upstream sets `description`, `wantedBy`, `wants = ["cups.service"]`, `after = ["cups.service"]`, `serviceConfig = { Type = "oneshot"; RemainAfterExit = true; }` and `script` — `wants`/`after` are list options (definitions concatenate), `serviceConfig` is an `attrsOf` with disjoint keys here, and upstream sets neither `unitConfig` nor `Restart`. So this merges rather than conflicts, and the resulting unit keeps its `After=cups.service`.

Placement is defensible: `config/profiles/pc/default.nix` is the only importer of this module and it also imports `../../modules/security/wireguard`, and crux doesn't import it at all (it configures its printer directly in `config/machines/crux/circinus.nix`), so there's no machine that picks up this ordering without the tunnel. Even if one appeared, `After=`/`Wants=` on a nonexistent unit is non-fatal in systemd. The module already hard-codes a single tunnel-only printer, so hard-coding the tunnel name next to it doesn't widen the coupling.

---

cprussin (commented):
Reviewed at `cbb9d76`. The diagnosis is right and the shape of the fix is right; I verified the parts that are easy to get wrong:

- **`wireguard-prussinnet.target` is the correct unit name** in the pinned nixpkgs (checked the module source at both `d407951` and `a9e6d84` — identical), and `useNetworkd` isn't set anywhere here, so the target actually exists.
- **`Restart=on-failure` is legal for `Type=oneshot`** — systemd refuses only `always`/`on-success`.
- **The added attrs merge with the nixpkgs-defined unit** rather than clobbering it, and the `mkIf` guard upstream is satisfied by this same file.
- **Nothing blocks boot**: `ensure-printers` is only `wantedBy=multi-user.target` with no `Before=`, so the retry loop runs beside the rest of the boot rather than delaying it.
- **Syntax/formatting** matches alejandra's output and the sibling files (2-space indent, `["…"]` with no inner padding, `serviceConfig`/`unitConfig` attrsets). I couldn't run `alejandra --check` — no nix in this environment — but there's nothing in the block that alejandra would reformat.

One substantive issue, inline on the `unitConfig` block: **the "~20 attempts / ~5 minutes" bound only holds if each `lpadmin` attempt fails in under ~16.5s.** systemd's start ratelimit resets its window when a start arrives more than `StartLimitIntervalSec` after the window opened, so slow failures (DNS to the tunnel's unreachable nameservers is 20–40s with glibc defaults — exactly the failure mode being defended against) make the limit never fire and the unit retry every 15s forever. `TimeoutStartSec = 10;` in `serviceConfig` makes the arithmetic deterministic. Everything else in the inline comments is affirmation or nits.

Two things worth considering separately, not blockers for this PR:

- **Nothing re-runs the unit when the tunnel comes up later.** Boot at a café, connect 20 minutes later, and the unit is permanently failed — same as today, so no regression, but it's the common laptop case. A re-trigger (a timer, with `RemainAfterExit` dropped since `lpadmin` is idempotent) addresses it in a way a longer burst can't.
- **`lyra` persists a long list of `tank/persisted-state/*` datasets** but not `/var/lib/cups`. Adding it would make the printer survive a boot outright, so a losing race would be cosmetic rather than leaving the machine with no printer. Complementary to this change, not a substitute — the unit would still fail.

---

config/modules/system/devices/printers/default.nix
cprussin: Fixed in `bc5e92b`: added `TimeoutStartSec = "10s"` and switched `RestartSec` to `"15s"`. Worst-case cycle is now 25s, so 20 starts land in ~475s and the limit actually trips.

Skipping the `reset-failed` sentence — keeping the comment short was an explicit ask. The re-trigger and persisting `/var/lib/cups` are worth doing but out of scope here.

---

config/modules/system/devices/printers/default.nix
```diff
@@ -12,4 +12,26 @@ _: {
...
+    wants = ["wireguard-prussinnet.target"];
+    after = ["wireguard-prussinnet.target"];
+    serviceConfig = {
+      Restart = "on-failure";
+      RestartSec = "15s";
+
+      # Without a bound on the attempt itself the retry limit below can never
+      # trip: a lookup against the tunnel's unreachable nameservers takes tens
+      # of seconds, and systemd restarts the ratelimit window whenever a start
```
cprussin: The start limit only bounds the retries if each failed attempt is fast — here it probably isn't.

systemd's rate limiter is a *fixed* window that resets whenever a start arrives more than `StartLimitIntervalSec` after the window opened (`ratelimit_below()`, `src/basic/ratelimit.c`), not a sliding one. So the limit trips only if all 20 starts fit inside 600s: spacing < 600/19 ≈ 31.6s, i.e. the attempt itself must fail in ≲16s on top of `RestartSec=15`. If it's slower, the window resets on every start, the counter never reaches the burst, and the unit retries forever.

The failure mode this PR targets is the slow one. `postSetup` (`config/modules/security/wireguard/default.nix`) points resolv.conf at the two wireguard-side nameservers, which are routed into the tunnel and black-holed until a handshake, and this box has no `systemd-resolved` (iwd with `Network.NameResolvingService = "resolvconf"`), so `lpadmin` blocks in glibc's stub resolver — `timeout:5` × `attempts:2` × 2 nameservers, i.e. tens of seconds — before it can even report failure. Worst case it hangs and is killed at `DefaultTimeoutStartSec` (90s). Either way spacing exceeds ~32s, so on a boot where the tunnel never comes up (laptop off-network) this retries every ~35–105s indefinitely instead of stopping after ~5 minutes.

Two ways to make the bound real:

- cap the attempt: `TimeoutStartSec` of ~10–15s keeps spacing under ~31s so the existing burst applies; or
- cap the total: `StartLimitIntervalSec = "infinity"`, documented in `systemd.unit(5)` as "limit the total number of start attempts, even if they happen at large time intervals" — 20 attempts regardless of spacing.

The `RestartSec = 15` / burst-20 arithmetic is right for the fast-failure case (starts at t=0…285s, the 21st denied); it's only the slow-failure case that escapes.

---

cprussin (commented):
Review of `cbb9d76`. The earlier thread on the unit name and the ordering-vs-retry premise still stands and I'm not repeating it; one new finding below, everything else checked out.

**Verified clean**

- **`Restart=on-failure` with `Type=oneshot` + `RemainAfterExit=true` is legal.** `systemd.service(5)` rejects only `always` and `on-success` for oneshot. NixOS's own `checkService` assertion (`nixos/lib/systemd-unit-options.nix`) also lists `on-failure` as valid, so this passes eval and systemd's own parser.
- **`StartLimitIntervalSec` / `StartLimitBurst` belong in `unitConfig`** (`[Unit]` section since systemd 229) and don't collide: `nixos/lib/systemd-lib.nix` emits those keys only `optionalAttrs (options.startLimitIntervalSec.isDefined)`, and nothing defines them here.
- **The merge into the nixpkgs-defined `ensure-printers` is clean.** `wants`/`after` are `listOf unitNameType` so they concatenate with upstream's `["cups.service"]`; `serviceConfig`/`unitConfig` are `attrsOf unitOption` merging per key, and `Restart`/`RestartSec` don't overlap upstream's `Type`/`RemainAfterExit`.
- **The retry can actually fire.** `makeJobScript` wraps `script` in `set -e`, so a failing `lpadmin` fails the unit rather than exiting 0.
- **Blast radius is right.** The printers module is only pulled in by `config/profiles/pc`, which also imports `config/modules/security/wireguard` — so the target always exists where this applies. `crux` is on `profiles/server` and sets `hardware.printers` directly in `circinus.nix`, so its own `ensure-printers` is untouched.
- **Shortened comment reads accurately.** `-m everywhere` really does make `lpadmin` fetch the queue's capabilities over IPP at unit-start time, and the two clauses still map onto the two mechanisms: "only works once the tunnel carries traffic" motivates the ordering, "that's a while after the tunnel's units finish" motivates the retry. It doesn't under-explain.
- **Formatting.** 2-space indent, no space inside list brackets, no trailing whitespace or tabs — matches the alejandra output in sibling files (`config/modules/system/users/default.nix`, `config/machines/crux/hardware.nix`). Nothing here that `alejandra --check` should reformat.

**Nit (non-blocking)**

NixOS has first-class `startLimitIntervalSec` / `startLimitBurst` options on `systemd.services.<name>`; they render to the same `[Unit]` keys but are `types.int`-checked rather than free-form `unitOption`. Worth preferring over `unitConfig` unless you're deliberately mirroring the systemd spelling.

Not approving, per my instructions — leaving the call to you.

---

config/modules/system/devices/printers/default.nix
cprussin: Took the second option in `e1c61c4`: `StartLimitIntervalSec = "infinity"` with the burst of 20. Preferred over `TimeoutStartSec` because it doesn't put a ceiling on a legitimately slow success, and it stretches the retry window in exactly the case where more patience helps — wifi that associates a few minutes into the boot.

---

PR #26, opened 2026-08-29
[crux] Fix the private server's unit

cprussin:
Debugging why `nix develop --command cli start-prod` is solid in the foreground but flaps under systemd. Three defects, all in the unit:

**Ordering against `import-tank` never existed.** `Requires=`/`After=` were in `[Install]`, where systemd knows no such keys and drops them with a warning. A home-manager *user* unit couldn't have expressed it from `[Unit]` either — a user manager has no `import-tank.service`, and `Requires=` on a nonexistent unit refuses the start. `/` is a tmpfs and `/home` comes off tank via `zfs mount -a`, so every boot raced the pool for `~/Private`.

**Nothing kept the user manager alive.** Lingering is off, so `user@1000.service` only exists between login and logout. `default.target` never came up at boot, and the server went down with the last session.

**The restart policy converted stumbles into outages.** `RestartSec` defaults to 100ms, so five restarts fit inside the limiter's 10s window and systemd then refuses to start the unit at all until `systemctl reset-failed`. Same trap #24 removed from `ensure-printers`. And `on-failure` ignores a clean exit, which for a server is just a quieter outage.

## The change

Makes it a system unit, matching every other service on this machine (`syncthing`, `immich-server`, `matrix`, `nvr`, …):

- `requires`/`after` on `import-tank.service` now actually hold, and — new — the unit releases the pool ahead of `preStop`'s `zpool export`, which nothing did before.
- `wantedBy = multi-user.target` starts it at boot with no login involved.
- `User` + `WorkingDirectory` replace the shell wrapper.
- `Restart = "always"`, `RestartSec = "10s"`: 10s spacing puts the default limiter out of reach, so no start-limit override is needed.

## What this buys you in monitoring: nothing

Worth stating plainly, because two review passes got it wrong before landing here. With `Restart=always` and the start limiter unreachable, **this unit has no path into a `failed` state at all.** A dependency failure finishes the *start job* with `JOB_DEPENDENCY` and leaves the unit `inactive (dead)` — it never appears in `systemctl --failed`. A `~/Private` that can never start just retries every 10s forever in `auto-restart`.

That is the accepted cost of never giving up on a server, and it is strictly better than the old behavior (which gave up permanently after five fast restarts). But the only evidence of an outage is the journal. If you want an alert, that's a separate change — `OnFailure=` has nothing to hang off here.

## Corrected during review

An earlier revision claimed the `/bin/sh -c 'cd ~/Private && …'` wrapper broke supervision (shell as MainPID) and that `~` resolved only by accident of `$HOME`. **Both were wrong**, and the commit message says so. NixOS's `/bin/sh` is bash, which execs the last command of a `-c` list — verified: `bash -c 'cd /tmp && sleep 4' &` yields a job whose PID *is* `sleep`, no shell remaining — so `nix` was already the main process. And bash's tilde expansion falls back to the passwd entry: `env -u HOME bash -c 'echo ~'` prints `/root`. Dropping the wrapper is still right, but only because `WorkingDirectory` is declarative and visible to `systemctl show`.

Also dropped after review: an explicit `startLimitIntervalSec = 0`. With `RestartSec = "10s"` the default 5-starts-per-10s limiter is already unreachable, so disabling it was a no-op.

## Ruled out

Memory pressure and the per-restart flake evaluation both looked like candidates and are not: the machine has 62GB and this workload doesn't approach it, and the re-evaluation is acceptable as-is.

Left alone deliberately: `Type = "simple"` reports the unit started before Next.js binds, so nginx 502s for the first few seconds of each start.

## Worth checking before deploying

- **The evidence gap.** Two of the three defects predict *"the server is absent"*, not *"it dies and restarts constantly"*. Chronic restarts imply the user manager **was** running — which a long-lived ssh/tmux session would do — leaving the restart policy as the defect that directly explains the symptom. `journalctl --user -u private -b` would settle which story is true. Nothing here has been observed on the box.
- **`SSH_AUTH_SOCK`.** A system unit gets `HOME`/`USER`/`LOGNAME`/`SHELL` from the user database and nothing else — no ssh agent, stdin on `/dev/null`. If the `~/Private` flake has a `git+ssh` input, the post-GC refetch runs `git fetch` with no agent and no way to prompt.
- **First deploy.** If a live login session is still running the old `~/.config/systemd/user/private.service`, it holds `:42042` and the new system unit will crash-loop until it's stopped.
- **`systemctl --user restart private` becomes `sudo systemctl restart private`.** `primary-user.sudo-cmds` whitelists only `journalctl -alf`, so that now prompts for a password.

## Testing

Untested on the box — no `nix` in this environment. CI is green, including `eval crux` (which instantiates the machine, so the new options typecheck) and `cli test format`.

### DIFF
--- a/config/machines/crux/private.nix
+++ b/config/machines/crux/private.nix
@@ -45,21 +45,62 @@ in {
 
   networking.firewall.interfaces.prussinnet.allowedTCPPorts = [42042];
 
-  primary-user.home-manager.systemd.user.services.private = {
-    Unit = {
-      Description = "Private";
-    };
+  # A system unit rather than a home-manager user unit.  The user unit couldn't
+  # express what it needed: `Requires=`/`After=` sat in `[Install]`, a section
+  # where systemd knows no such keys and drops them, and a user unit can't order
+  # itself against a system unit like `import-tank` even from `[Unit]`.  Nor
+  # would it stay up -- without lingering the user manager only exists between
+  # login and logout, so `default.target` never came up at boot and every unit
+  # under it died with the last session.
+  systemd.services.private = {
+    description = "Private";
+    wantedBy = ["multi-user.target"];
 
-    Service = {
+    # `/` is a tmpfs and `/home` comes from tank, so the checkout doesn't exist
+    # until `import-tank` has run `zfs mount -a`.  `requires` propagates that
+    # unit's stop to this one; `after` is what orders this unit's stop ahead of
+    # the `zpool export` in its `preStop`, which nothing released the pool for
+    # before.
+    #
+    # `network-online.target` isn't for the server, which needs no address to
+    # exist before it binds -- nginx doesn't wait on it either.  It's for `nix
+    # develop`, which re-evaluates the flake on every start and, after the
+    # weekly `nix.gc`, may have to refetch to do it.
+    requires = ["import-tank.service"];
+    after = [
+      "import-tank.service"
+      "network-online.target"
+    ];
+    wants = ["network-online.target"];
+
+    # Nix still shells out to `git` for some flake refs, and it does that while
+    # evaluating -- with this PATH, before the devShell exists.  (`nix develop`
+    # then prepends the devShell's PATH rather than replacing it, so this stays
+    # reachable as the tail.)
+    path = [pkgs.git];
+
+    serviceConfig = {
       Type = "simple";
-      ExecStart = "/bin/sh -c 'cd ~/Private && ${pkgs.nix}/bin/nix develop --command cli start-prod'";
-      Restart = "on-failure";
-    };
+      User = config.primary-user.name;
 
-    Install = {
-      WantedBy = ["default.target"];
-      Requires = ["import-tank.service"];
-      After = ["import-tank.service"];
+      # `WorkingDirectory` rather than the old `/bin/sh -c 'cd ~/Private && ...'`.
+      # Not for supervision: NixOS's `/bin/sh` is bash, which execs the last
+      # command of a `-c` list, so `nix` was already the main process either way.
+      # It's that the directory is declared where `systemctl show` reports it,
+      # instead of buried in a string only a shell can read.
+      WorkingDirectory = "${config.primary-user.home}/Private";
+      ExecStart = "${pkgs.nix}/bin/nix develop --command cli start-prod";
+
+      # A server that stops serving is a failure however it exited, so
+      # `on-failure` was wrong: a clean exit left the port dead until someone
+      # noticed.  `RestartSec` is the other half of it -- at the 100ms default,
+      # five restarts fit inside the limiter's 10s window and systemd then
+      # refuses to start the unit at all until `systemctl reset-failed`.  Spaced
+      # 10s apart the limiter can't be reached, so there's nothing to disable.
+      # Nothing surfaces as a `failed` unit either -- a dependency failure fails
+      # the start job, not the unit -- so an outage shows up in the journal.
+      Restart = "always";
+      RestartSec = "10s";
     };
   };
 }


config/machines/crux/private.nix
```diff
@@ -45,21 +45,61 @@ in {
...
-      Requires = ["import-tank.service"];
-      After = ["import-tank.service"];
+      # A server that stops serving is a failure however it exited, so
+      # `on-failure` was wrong: a clean exit left the port dead until someone
+      # noticed.  `RestartSec` is the other half of it -- at the 100ms default,
+      # five restarts fit inside the limiter's 10s window and systemd then
+      # refuses to start the unit at all until `systemctl reset-failed`.  Spaced
+      # 10s apart the limiter can't be reached, so there's nothing to disable:
+      # the unit still lands in `failed` if a dependency fails, which is the one
+      # case worth seeing.
```
cprussin: The first half is right, the consolation clause isn't.

Verified: with `RestartSec=10s` the limiter genuinely can't be reached. `ratelimit_below()` resets the window whenever `now - begin > interval`, so at ≥10s spacing the counter tops out at 2 (only if a start lands at *exactly* the 10s boundary) against a burst of 5. So "spaced 10s apart the limiter can't be reached, so there's nothing to disable" is exact.

But "the unit still lands in `failed` if a dependency fails" is not how systemd behaves. When `import-tank.service` fails, `job_fail_dependencies()` finishes *this unit's start job* with `JOB_DEPENDENCY` — the job fails, the unit does not change state. It stays `inactive (dead)`, never shows in `systemctl --failed`, and `JOB_DEPENDENCY` isn't even counted in the manager's `n_failed_jobs`. systemd's own source says so at that site: "*Trigger `OnFailure=` dependencies manually here. We need to do that because a failed job might not cause a unit state change.*" Contrast the start-limit path, which really does set failed state (`service.c`: `service_enter_dead(s, SERVICE_FAILURE_START_LIMIT_HIT, false)`).

Which means: with the limiter unreachable and `Restart=always`, this unit has essentially no path into `failed` at all. A dependency failure leaves it dead-and-quiet; a `WorkingDirectory` that isn't there retries every 10s forever in `auto-restart`. That's not an argument against the config — it's the same tradeoff the previous revision stated correctly ("the cost is that a checkout that can never start retries quietly forever ... it shows up only in the journal"). The rewrite replaced a true statement of the cost with a false claim of a safety net.

Suggest ending the comment at the accurate part, e.g.:

```
      # ... Spaced
      # 10s apart the limiter can't be reached, so there's nothing to disable.
      # Nothing surfaces as a `failed` unit either -- a dependency failure fails
      # the start job, not the unit -- so an outage shows up in the journal.
```

---

config/machines/crux/private.nix
```diff
@@ -45,21 +45,61 @@ in {
...
+    after = [
+      "import-tank.service"
+      "network-online.target"
+    ];
+    wants = ["network-online.target"];
 
-    Service = {
+    # Nix still runs `git` for some flake refs.  It's worth being explicit
+    # because `nix develop` *prepends* the devShell's PATH rather than replacing
+    # it, so whatever is here stays reachable as the tail.
```
cprussin: Both facts here check out, but the "because" joining them doesn't.

- Nix does still shell out to `git`: `readHead()` in `src/libfetchers/git.cc` runs `git ls-remote --symref`, `storeCachedHead()` runs `git symbolic-ref`, and the `putFile` path runs `check-ignore`/`add`/`commit`.
- `nix develop` does prepend: `makeRcScript()` saves `PATH` and `XDG_DATA_DIRS` and re-emits `PATH="$PATH${nix_saved_PATH:+:$nix_saved_PATH}"` — devShell first, unit PATH as the tail — and that same rc script is used for `--command`, not just the interactive shell.

The problem is the causal link. `git` has to be here because `nix` execs it *while evaluating the flake*, using the unit's own PATH, before a devShell environment exists at all. That would be equally true if `nix develop` clobbered PATH. The prepend is a true but separate bonus (git stays reachable inside the shell too), so it can't be the reason it's "worth being explicit".

Suggest something like:

```
    # Nix still shells out to `git` for some flake refs, and it does that while
    # evaluating -- with this PATH, before the devShell exists.  (`nix develop`
    # then prepends the devShell's PATH rather than replacing it, so this stays
    # reachable as the tail.)
```

---

cprussin (commented):
Third pass on `769a65e`. Two comment-accuracy findings, one substantive; the config itself I believe is now correct.

**Verified good**

- **Limiter unreachable at `RestartSec=10s`.** `ratelimit_below()` starts a new window whenever `now - begin > interval`, so with starts ≥10s apart the counter never exceeds 2 of the burst of 5. Dropping `startLimitIntervalSec = 0` is a true no-op, and the NixOS module only emits `StartLimitIntervalSec=` when the option is defined, so nothing is left behind. I could not find a restart path that escapes `RestartSec`: auto-restarts are always delayed by it, `Requires=` propagates stop (not restart) from `import-tank`, and deploy-driven restarts are one-at-a-time.
- **Dropping `Group = "users"` is safe and equivalent.** `modules/nixos/primary-user.nix` sets `isNormalUser = true`, and NixOS defaults `users.users.<name>.group` to `"users"` for normal users; systemd uses the user's primary group when `Group=` is unset. Same gid either way, now not duplicated.
- **The `WorkingDirectory` rationale.** Confirmed empirically on bash 5.2 that `-c 'cd X && cmd'` execs `cmd` in place — including when bash is invoked as `sh` — so "`nix` was already the main process either way" holds, and no `environment.binsh` override in this repo changes `/bin/sh` away from bash.
- The `network-online.target` and `nix.gc` premises still line up with the rest of the tree (`nix.gc.dates = "weekly"`, `import-tank`'s `preStop = zpool export tank`, scripted networking where `dhcpcd` is `before`/`wantedBy` `network-online.target`).

**Not verified** — I have no NixOS host or nix store here, so everything above is from systemd/nix/nixpkgs sources at the pinned or current revisions rather than from runtime; and I can't see the `~/Private` flake, so "some flake refs" is taken generically.

---

cprussin (commented):
Fourth pass, head `aa3632b`. No findings — ship it.

Scope check: `git diff 769a65e aa3632b` is two comment hunks in `config/machines/crux/private.nix` and nothing else. No code, no option changes.

Both rewrites are accurate on their own terms, not just faithful transcriptions:

**`path = [pkgs.git]` (~line 78).** The causal chain is now right. Nix does exec `git` for some flake refs, and it does so during evaluation, in the `nix develop` process itself — which inherits the unit's `PATH`, since the devShell environment only exists for the `--command` child. So `pkgs.git` here is load-bearing, not decorative, and the old "it's worth being explicit because *`nix develop` prepends*" inversion is gone. The prepend fact is still true (`nix develop` saves the outer `PATH` and re-appends it after sourcing the dev env, rather than replacing it) and is now correctly demoted to a parenthetical bonus rather than the justification.

**`Restart`/`RestartSec` (~line 96).** The false claim is gone and the replacement is correct: a `Requires=` dependency failure finishes the start job with `JOB_DEPENDENCY`, leaving the unit `inactive (dead)`, never `failed`. Dependency *stop* propagation to an already-running unit likewise deactivates rather than fails it. Combined with `Restart=always` and a start limiter that `RestartSec=10s` keeps out of reach, there is genuinely no path to a `failed` unit here, so "an outage shows up in the journal" is the honest statement of what monitoring you actually get. The preceding sentences already carry the limiter arithmetic (100ms default × 5 starts inside the 10s window), so the new clause doesn't overclaim as the sole reason.

Nothing else in the file regressed; the surrounding comments are unchanged from the previously reviewed head and remain accurate.

---

PR #27, opened 2026-08-29
[crux] Add the credit card offer adder, and ntfy to answer its code prompts

cprussin:
Runs `credit-card-offer-adder` on crux twice a day across both Amex and both Chase accounts, plus the ntfy server its one-time-code ladder needs.

## Shape

| Piece | Where |
|---|---|
| `services.offer-adder` + its credentials key | `config/machines/crux/offer-adder.nix` |
| `services.ntfy-sh`, its nginx vhost, its auth key | `config/machines/crux/ntfy.nix` |
| `getOfferAdderCredentials`, `getNtfySecrets` | `lib/passwords.nix` |
| flake input (no `nixpkgs.follows`) + lock | `flake.nix`, `flake.lock` |
| `ntfy.internal.prussin.net` | `config/machines/crux/dns.nix` |
| profile dataset excluded from run-backup | `config/machines/crux/backup.nix` |

## Decisions

**ntfy rather than a mailbox.** Neither issuer offers authenticator-app enrollment, so every account will be challenged with a code no machine can compute. IMAP was the alternative; it needs a distinct address on file at each bank and a shared mailbox risks one run reading the other person's code. ntfy is one tap on a phone instead. Every account is `codeSources = ["ntfy"]` — `"totp"` is *not* a preference that falls through to the rung behind it, so it goes in only alongside a real `TOTP Secret` field in pass.

**ntfy's auth database is provisioned from the environment, not the CLI.** `NTFY_AUTH_USERS` / `NTFY_AUTH_ACCESS` / `NTFY_AUTH_TOKENS` come from a colmena key built out of pass, so there is no `ntfy user add` step to remember and nothing to restore if `/var/lib/ntfy-sh` is lost — which matters, because that directory stays on the tmpfs root by design. The one consequence is that a token minted at runtime through `/v1/account/token` does not survive a reboot while the provisioned one does, so the phones should log in with username and password. `auth-default-access: deny-all` — every topic is private and the ACLs are the only way in. Password hashes are `mkpasswd -m bcrypt -R 10`.

**The offer adder talks to ntfy over loopback.** nginx terminates TLS for the phones; the service, on the same box, uses `http://127.0.0.1:2586` directly. That keeps the internal CA out of bun's certificate store, which does not read the system trust store the way the rest of the host does.

**The browser profiles get their own tank dataset.** `/` is a tmpfs and `/var/lib` is not otherwise covered, so `tank/persisted-state/offer-adder` mounts at `/var/lib/offer-adder` natively under `import-tank` — which the unit now `Requires`, so a catch-up run before the pool is imported fails loudly instead of writing profiles somewhere `zfs mount -a` then shadows. It holds live bank sessions and a restored stale copy earns a challenge on every account anyway, so neither backup takes it: excluded from run-backup's `zfs send -R` (opt-out), and the borg job is opt-in on `net.prussin:backup`, which inherits — so confirm no ancestor sets it rather than assuming an unset child is enough.

**`sealed = false`.** colmena hands over the plaintext document, so `LoadCredential` rather than `LoadCredentialEncrypted`. systemd still stages it in the per-unit tmpfs; what is given up is the TPM binding. The credentials key uses `destDir = "/secrets"` — `/run/keys` is a tmpfs colmena only refills on deploy.

**09:17 and 19:17, not 03:17 and 15:17.** ntfy waits five minutes for a reply. A challenge at 3am reaches nobody and costs that account its run.

**No secret in argv, and none on stdout on failure.** `ps` is world-readable on the deploying host, so both key commands pass values to `jq` through the environment; `getOfferAdderCredentials` assembles the document into a variable so a failed lookup emits nothing at all rather than a partial document, and the ntfy token is read by a plain assignment ahead of the loop — `VAR="$(...)" prog` is the one construct `set -e` cannot see.

## Verification

No nix binary in the session environment initially, so this was checked by building one.

**Against the versions this lock actually pins**, which is not the same as the branch `flake.nix` names — the lock's root `nixpkgs` records `original.ref = "nixos-unstable"` (pre-existing, untouched here), so crux gets `d407951`:

- **ntfy 2.25.0**, not 26.05's 2.26.0. Started with the generated `server.yml` and the generated environment file: anonymous publish denied (403), off-ACL topic denied (403), token publish accepted, human basic-auth accepted, wrong password rejected (401), `/v1/account/token` works for provisioned users so the phone app can log in. Then end to end through the app's own code — `ntfyCodeSource` published the alert with the provisioned token, streamed the reply topic, and picked up a code published back by the "phone" user. Verified identically on 2.26.0. `eval crux` cannot catch a version problem here, because the provisioning goes through freeform env vars.
- **playwright 1.61.1.** At `9fbb54b`, the nixpkgs this lock pins for the offer-adder, `playwright-driver` is 1.61.1, and the offer-adder's catalog pins `"playwright": "1.61.1"` bare. That exact match is the condition `docs/DEPLOYMENT.md` requires or the first `launchPersistentContext` fails.
- `services.ntfy-sh.environmentFile` exists at the locked rev, and `jq` is 1.8.1 (the object-value syntax is parenthesised, so 1.7 would work too).

**Everything else:**

- `alejandra --check`, `statix check`, `deadnix --fail` — all clean.
- Both new machine modules evaluated against a real nixpkgs tree (with a stub for colmena's `deployment.keys`), asserting the generated `offer-adder.service`, `offer-adder.timer`, `ntfy-sh.service`, `server.yml` and `offers.config.json`. No failed assertions. CI's own `eval crux` job instantiates the whole machine.
- The generated `offers.config.json` parsed with the app's own `loadConfig`, and every account's ladder built with the app's own `buildCodeSource` against a credentials document carrying no TOTP secrets — the state the deployment is actually in.
- Both key commands run against a fake pass store, including a password containing `$`, `'` and `"`. Every failure path — missing `Token`, nonexistent ntfy entry, missing `Username`, odd argument count, nonexistent bank entry — exits 1 with empty stdout.
- The lock adds 8 nodes, all under `credit-card-offer-adder`, and re-pins nothing: walking the resolved input graph from `root`, every pre-existing path resolves to a byte-identical locked source. `bun2nix.nixpkgs` follows the offer-adder's, so there is exactly one extra nixpkgs.

## Before the first run

```sh
# -p because tank/persisted-state is lyra's naming; nothing shows one on crux yet.
# It still applies -o mountpoint= to the leaf, only the ancestors go propertyless.
zfs create -p -o mountpoint=/var/lib/offer-adder tank/persisted-state/offer-adder
zfs get -s local,inherited net.prussin:backup tank tank/persisted-state   # expect no `true`
```

(An earlier revision of this description claimed `run-backup` had to wait for the dataset. It does not — `zfs send -X` validates its argument for name syntax only and never resolves it, so the exclusion is inert until the dataset exists. Corrected on the `backup.nix` thread.)

## Not included

Nothing in `credit-card-offer-adder` needed to change. Its NixOS module already supports everything this deployment needs.

### DIFF
--- a/config/machines/crux/backup.nix
+++ b/config/machines/crux/backup.nix
@@ -49,7 +49,11 @@
 
     echo
     echo -e "\e[1;37mSending incremental data...\e[0m"
-    zfs send -R -X tank/Cache,tank/Data/cprussin/Private,tank/Data/cprussin/Scratch,tank/Data/frigate -I tank@external-backup-$LAST_SNAP_DATE tank@external-backup-$TODAY | zfs recv -Fdu tank-backup
+    # -X is opt-out, unlike the borg job's net.prussin:backup, so anything new
+    # under tank lands on the external disk by default.  offer-adder holds live
+    # bank session cookies, and restoring a stale copy earns a challenge on
+    # every account anyway, so replicating it is all cost.
+    zfs send -R -X tank/Cache,tank/Data/cprussin/Private,tank/Data/cprussin/Scratch,tank/Data/frigate,tank/persisted-state/offer-adder -I tank@external-backup-$LAST_SNAP_DATE tank@external-backup-$TODAY | zfs recv -Fdu tank-backup
 
     echo
     echo -e "\e[1;37mScrubbing tank-backup...\e[0m"
--- a/config/machines/crux/ntfy.nix
+++ b/config/machines/crux/ntfy.nix
@@ -1,0 +1,113 @@
+# ntfy is the offer adder's escape hatch: when Amex or Chase insists on a
+# one-time code, the run pushes a notification here and waits for a phone to
+# publish the digits back. It is wireguard-only and its whole auth database --
+# users, ACLs and the service account's token -- is provisioned from pass at
+# start, so nothing about it is imperative.
+{
+  config,
+  pkgs,
+  ...
+}: let
+  passwords = pkgs.callPackage ../../../lib/passwords.nix {};
+  network = pkgs.callPackage ../../../lib/network.nix {};
+  host = "ntfy.internal.prussin.net";
+  # Loopback only: nginx terminates TLS for the phones, and the offer adder
+  # running on this same box talks to it directly, which keeps the internal CA
+  # out of bun's certificate store entirely.
+  listenAddress = "127.0.0.1:2586";
+in {
+  deployment.keys = {
+    ntfy-secrets = {
+      keyCommand =
+        passwords.getNtfySecrets "offers-*"
+        "Connor/Infrastructure/ntfy/offer-adder"
+        [
+          "Connor/Infrastructure/ntfy/connor"
+          "Shauna/Infrastructure/ntfy/shauna"
+        ];
+      destDir = "/secrets";
+    };
+
+    "${host}.crt" = {
+      inherit (config.users.users.nginx) group;
+      keyCommand = passwords.getFullPassword "Connor/Infrastructure/ssl/${host}/cert";
+      user = config.users.users.nginx.name;
+    };
+
+    "${host}.key" = {
+      inherit (config.users.users.nginx) group;
+      keyCommand = passwords.getFullPassword "Connor/Infrastructure/ssl/${host}/key";
+      user = config.users.users.nginx.name;
+    };
+  };
+
+  networking.firewall.interfaces.prussinnet.allowedTCPPorts = [80 443];
+
+  services = {
+    ntfy-sh = {
+      enable = true;
+      environmentFile = config.deployment.keys.ntfy-secrets.path;
+      settings = {
+        base-url = "https://${host}";
+        listen-http = listenAddress;
+        behind-proxy = true;
+        # Every topic is private; the ACLs in the environment file are the only
+        # way in.  The auth database is rebuilt from that file on every start,
+        # so its StateDirectory is deliberately left on the tmpfs root: nothing
+        # in it needs to survive a reboot except what the environment already
+        # carries.  The one consequence is that a token minted at runtime
+        # through /v1/account/token does not come back, so the phones should
+        # authenticate with their username and password.
+        auth-default-access = "deny-all";
+      };
+    };
+
+    nginx = {
+      enable = true;
+      recommendedTlsSettings = true;
+      recommendedOptimisation = true;
+      recommendedGzipSettings = true;
+      recommendedProxySettings = true;
+      virtualHosts."${host}" = {
+        listenAddresses = ["[${network.wireguard6.crux.address}]" "${network.wireguard4.crux.address}"];
+        sslCertificate = config.deployment.keys."${host}.crt".path;
+        sslCertificateKey = config.deployment.keys."${host}.key".path;
+        forceSSL = true;
+        http2 = true;
+        locations."/" = {
+          proxyPass = "http://${listenAddress}";
+          proxyWebsockets = true;
+          # Subscribers hold a stream open until something is published, so
+          # buffering it would defeat the point entirely.  The read timeout is
+          # only belt and braces -- ntfy keepalives every 45s, inside nginx's
+          # 60s default -- but it is set past the five minutes a run waits for
+          # a code, so a phone that misses a keepalive still holds the stream
+          # across a whole request.
+          extraConfig = ''
+            proxy_buffering off;
+            proxy_request_buffering off;
+            proxy_read_timeout 6m;
+          '';
+        };
+      };
+    };
+  };
+
+  systemd.services = {
+    ntfy-sh = {
+      after = ["ntfy-secrets-key.service"];
+      requires = ["ntfy-secrets-key.service"];
+    };
+
+    nginx = {
+      after = [
+        "${host}.crt-key.service"
+        "${host}.key-key.service"
+      ];
+      requires = [
+        "${host}.crt-key.service"
+        "${host}.key-key.service"
+      ];
+    };
+  };
+}
--- a/config/machines/crux/offer-adder.nix
+++ b/config/machines/crux/offer-adder.nix
@@ -1,0 +1,108 @@
+# Twice a day, walk both Amex and both Chase accounts and add every offer the
+# issuers are handing out. The bank logins live in pass and are assembled into
+# the one JSON document the service reads; systemd stages it in a tmpfs only
+# this unit can see, so nothing lands in the environment or in `systemctl show`.
+{
+  config,
+  pkgs,
+  ...
+}: let
+  passwords = pkgs.callPackage ../../../lib/passwords.nix {};
+in {
+  deployment.keys.offer-adder-credentials = {
+    keyCommand =
+      passwords.getOfferAdderCredentials "Connor/Infrastructure/ntfy/offer-adder"
+      {
+        connor-amex = "Connor/Financial/American Express";
+        connor-chase = "Connor/Financial/Chase";
+        shauna-amex = "Shauna/Financial/American Express";
+        shauna-chase = "Shauna/Financial/Chase";
+      };
+    # /run/keys is a tmpfs that colmena refills on deploy, so a reboot without
+    # an `upload-keys` would silently break every run until the next deploy.
+    destDir = "/secrets";
+  };
+
+  services.offer-adder = {
+    enable = true;
+    credentialFile = config.deployment.keys.offer-adder-credentials.path;
+    # colmena hands over the plaintext document; systemd still stages it in the
+    # per-unit tmpfs, it just has no TPM sealing to undo first.
+    sealed = false;
+    # A challenge that reaches nobody costs the account its run -- ntfy waits
+    # five minutes for a reply -- so fire when somebody is awake rather than at
+    # the small hours the module defaults to.
+    onCalendar = "*-*-* 09:17,19:17:00";
+    settings = {
+      # Every account is `["ntfy"]` because neither issuer offers
+      # authenticator-app enrollment: a challenge always ends up on a phone.
+      # `"totp"` is not a preference that falls through to the rung behind it
+      # -- the ladder resolves the secret up front and fails the account for
+      # the whole run when there is none -- so it goes in only for an account
+      # actually enrolled, alongside a `TOTP Secret` field in its pass entry.
+      ntfy = {
+        server = "http://${config.services.ntfy-sh.settings.listen-http}";
+        alertTopic = "offers-alerts";
+        replyTopic = "offers-codes";
+      };
+      accounts = [
+        {
+          id = "connor-amex";
+          label = "Connor · Amex";
+          issuer = "amex";
+          senderHints = ["americanexpress" "american express"];
+          codeSources = ["ntfy"];
+        }
+        {
+          id = "connor-chase";
+          label = "Connor · Chase";
+          issuer = "chase";
+          senderHints = ["chase.com" "chase"];
+          codeSources = ["ntfy"];
+        }
+        {
+          id = "shauna-amex";
+          label = "Shauna · Amex";
+          issuer = "amex";
+          senderHints = ["americanexpress" "american express"];
+          codeSources = ["ntfy"];
+        }
+        {
+          id = "shauna-chase";
+          label = "Shauna · Chase";
+          issuer = "chase";
+          senderHints = ["chase.com" "chase"];
+          codeSources = ["ntfy"];
+        }
+      ];
+    };
+  };
+
+  # The service's StateDirectory holds one Chromium profile per account, and
+  # those profiles are the whole reason the banks stop challenging every login.
+  # `/` here is a tmpfs, so they live on `tank/persisted-state/offer-adder`,
+  # created with `mountpoint=/var/lib/offer-adder` and mounted natively by
+  # `zfs mount -a` like the rest of crux's service state -- not declared in
+  # `fileSystems`, which is for the tank-fast datasets.  `StateDirectory`
+  # chowns the dataset root to the service user on each start, so nothing
+  # here has to.  It is excluded from `run-backup` in ./backup.nix, and the
+  # borg job is opt-in on `net.prussin:backup` -- a user property, so confirm
+  # no ancestor sets it (`zfs get -s local,inherited net.prussin:backup tank
+  # tank/persisted-state`) rather than assuming an unset child is enough.
+  systemd.services.offer-adder = {
+    # `import-tank` is what mounts that dataset; without it the timer's
+    # catch-up run at boot writes profiles to the tmpfs root, and the later
+    # `zfs mount -a` silently shadows what it wrote.  ntfy is only wanted:
+    # requiring it would refuse the whole unit on a systemd dependency error,
+    # where wanting it lets the run happen and report which accounts could not
+    # get a code and why -- and a run only needs ntfy for an account the bank
+    # actually challenges, which most runs are not.
+    after = [
+      "offer-adder-credentials-key.service"
+      "import-tank.service"
+      "ntfy-sh.service"
+    ];
+    requires = ["offer-adder-credentials-key.service" "import-tank.service"];
+    wants = ["ntfy-sh.service"];
+  };
+}
--- a/flake.nix
+++ b/flake.nix
@@ -12,6 +12,10 @@
 
     colmena.url = "github:zhaofengli/colmena";
 
+    # Deliberately not `nixpkgs.follows`: it pins playwright to the exact
+    # chromium its own nixpkgs ships, and a mismatched pair refuses to launch.
+    credit-card-offer-adder.url = "github:cprussin/credit-card-offer-adder";
+
     dircolors-solarized = {
       url = "github:seebi/dircolors-solarized";
       flake = false;
@@ -44,6 +48,7 @@
     home-manager,
     nixos-hardware,
     colmena,
+    credit-card-offer-adder,
     ...
   } @ flake-inputs: let
     password-utils-overlay = final: _: {
@@ -130,6 +135,9 @@
 
       crux = mkMachine {
         targetHost = "crux";
+        extraModules = [
+          credit-card-offer-adder.nixosModules.default
+        ];
       };
     };
   in
--- a/lib/passwords.nix
+++ b/lib/passwords.nix
@@ -1,10 +1,14 @@
-{pkgs}: let
+{
+  pkgs,
+  lib,
+}: let
   pass = "${pkgs.pass}/bin/pass";
   head = "${pkgs.coreutils}/bin/head";
   base64 = "${pkgs.coreutils}/bin/base64";
   grep = "${pkgs.gnugrep}/bin/grep";
   sed = "${pkgs.gnused}/bin/sed";
   mkpasswd = "${pkgs.mkpasswd}/bin/mkpasswd";
+  jq = "${pkgs.jq}/bin/jq";
 
   getFullPassword = pkgs.writeShellScriptBin "getFullPassword" ''
     set -euo pipefail
@@ -56,6 +60,95 @@
     fi
   '';
 
+  # The offer adder wants one JSON document holding every bank login, so this
+  # walks `<account id> <pass path>` pairs and assembles it. Each entry's first
+  # line is the password; `Username` and the optional `TOTP Secret` are fields.
+  # Nothing secret is ever passed in argv -- `ps` is world-readable -- so the
+  # values reach jq through the environment instead.
+  getOfferAdderCredentials = pkgs.writeShellScriptBin "getOfferAdderCredentials" ''
+    set -euo pipefail
+    # A plain assignment, and ahead of the loop: `VAR="$(...)" prog` is the one
+    # construct `set -e` cannot see, so a lookup that fails in an assignment
+    # prefix would hand jq an empty token and exit 0.  Doing it first also
+    # fails a bad ntfy entry before any bank password is read.
+    ntfyToken="$(${getPasswordField}/bin/getPasswordField "$1" "Token")"
+    shift
+    # Assembled into a variable rather than straight to stdout: `jq -s` on the
+    # right of the pipe would otherwise emit a partial document before the
+    # non-zero status of a failed lookup could stop it.
+    document="$(
+      {
+        while [ "$#" -gt 0 ]; do
+          if [ "$#" -lt 2 ]; then
+            echo "getOfferAdderCredentials: account \"$1\" has no pass path" >&2
+            exit 1
+          fi
+          entry="$(${pass} show "$2")"
+          # `TOTP Secret` is absent on every account today -- neither issuer
+          # offers authenticator-app enrollment -- and the credentials schema
+          # makes it optional, so it is emitted only when present.  An absent
+          # `Username` is always a malformed entry, though, and `sed` would
+          # otherwise hand over an empty string that deploys clean and fails
+          # at the bank sign-in twelve hours later.
+          username="$(printf '%s\n' "$entry" | ${sed} -n 's/^Username: //p')"
+          if [ -z "$username" ]; then
+            echo "getOfferAdderCredentials: no Username field in \"$2\"" >&2
+            exit 1
+          fi
+          OFFER_ADDER_PASSWORD="$(printf '%s\n' "$entry" | ${head} -n 1)" \
+          OFFER_ADDER_USERNAME="$username" \
+          OFFER_ADDER_TOTP="$(printf '%s\n' "$entry" | ${sed} -n 's/^TOTP Secret: //p')" \
+            ${jq} -cn --arg id "$1" '{
+              key: $id,
+              value: (
+                {
+                  username: $ENV.OFFER_ADDER_USERNAME,
+                  password: $ENV.OFFER_ADDER_PASSWORD,
+                }
+                + (
+                  if $ENV.OFFER_ADDER_TOTP == "" then {}
+                  else {totpSecret: $ENV.OFFER_ADDER_TOTP}
+                  end
+                )
+              ),
+            }'
+          shift 2
+        done
+      } | OFFER_ADDER_NTFY_TOKEN="$ntfyToken" \
+        ${jq} -s '{ntfyToken: $ENV.OFFER_ADDER_NTFY_TOKEN, accounts: from_entries}'
+    )"
+    printf '%s\n' "$document"
+  '';
+
+  # ntfy provisions its users, their ACLs and their access tokens from the
+  # environment at startup, so the whole auth database is derivable from pass.
+  # Each entry's first line is the password and `Username` names the ntfy user;
+  # the first path is the service account, whose `Token` field the offer adder
+  # authenticates with. That field must be `ntfy token generate` output --
+  # `tk_` and 29 more characters -- because ntfy validates provisioned tokens
+  # at startup and refuses to serve at all when one is malformed.
+  getNtfySecrets = pkgs.writeShellScriptBin "getNtfySecrets" ''
+    set -euo pipefail
+    topic="$1"
+    servicePath="$2"
+    # Drop only the topic: the service account is a user like any other, it
+    # just also carries the token.
+    shift
+    users=""
+    access=""
+    for path in "$@"; do
+      username="$(${getPasswordField}/bin/getPasswordField "$path" "Username")"
+      hash="$(${pass} show "$path" | ${mkpasswd} -m bcrypt -R 10 -s)"
+      users="''${users:+$users,}$username:$hash:user"
+      access="''${access:+$access,}$username:$topic:rw"
+    done
+    # Single quotes: systemd reads them literally, and a bcrypt hash is full of
+    # dollar signs.
+    echo "NTFY_AUTH_USERS='$users'"
+    echo "NTFY_AUTH_ACCESS='$access'"
+    echo "NTFY_AUTH_TOKENS='$(${getPasswordField}/bin/getPasswordField "$servicePath" "Username"):$(${getPasswordField}/bin/getPasswordField "$servicePath" "Token")'"
+  '';
+
   getImmichSecrets = pkgs.writeShellScriptBin "getImmichSecrets" ''
     set -euo pipefail
     echo "DB_PASSWORD=$(${getPassword}/bin/getPassword "$1")"
@@ -80,6 +173,8 @@ in {
       getVaultwardenSecrets
       getGmailNewMailCounterEnvFile
       getImmichSecrets
+      getNtfySecrets
+      getOfferAdderCredentials
     ];
   };
   getPassword = name: ["getPassword" name];
@@ -97,4 +192,9 @@ in {
     );
   getGmailNewMailCounterEnvFile = name: ["getGmailNewMailCounterEnvFile" name];
   getImmichSecrets = db: ["getImmichSecrets" db];
+  getNtfySecrets = topic: servicePath: humanPaths:
+    ["getNtfySecrets" topic servicePath] ++ humanPaths;
+  getOfferAdderCredentials = ntfyPath: accounts:
+    ["getOfferAdderCredentials" ntfyPath]
+    ++ lib.concatLists (lib.mapAttrsToList (id: path: [id path]) accounts);
 }
--- a/config/machines/crux/default.nix
+++ b/config/machines/crux/default.nix
@@ -13,7 +13,9 @@
     ./library.nix
     ./libvirtd.nix
     ./matrix.nix
+    ./ntfy.nix
     ./nvr.nix
+    ./offer-adder.nix
     ./photos.nix
     ./powerpanel.nix
     ./private.nix
--- a/config/machines/crux/dns.nix
+++ b/config/machines/crux/dns.nix
@@ -9,7 +9,7 @@
     lib.concatStringsSep "\n" (
       map
       (wgNet: ''
-        ${wgNet.crux.address} library.internal.prussin.net home-assistant.internal.prussin.net matrix.internal.prussin.net passwords.internal.prussin.net circinus.internal.prussin.net eyes.internal.prussin.net photos.internal.prussin.net private.internal.prussin.net
+        ${wgNet.crux.address} library.internal.prussin.net home-assistant.internal.prussin.net matrix.internal.prussin.net ntfy.internal.prussin.net passwords.internal.prussin.net circinus.internal.prussin.net eyes.internal.prussin.net photos.internal.prussin.net private.internal.prussin.net
         ${
           lib.concatStringsSep "\n" (
             lib.mapAttrsToList
--- a/flake.lock
+++ b/flake.lock
@@ -1,5 +1,29 @@
 {
   "nodes": {
+    "bun2nix": {
+      "inputs": {
+        "flake-parts": "flake-parts",
+        "nixpkgs": [
+          "credit-card-offer-adder",
+          "nixpkgs"
+        ],
+        "systems": "systems_2",
+        "treefmt-nix": "treefmt-nix"
+      },
+      "locked": {
+        "lastModified": 1784665499,
+        "narHash": "sha256-9BMxlTxCCDAeoNLtb1a/st7udtTIJep+wpUzquA29VU=",
+        "owner": "nix-community",
+        "repo": "bun2nix",
+        "rev": "0f2a1f0b6f42cebe3b149bf62d38754c5e0e9729",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-community",
+        "repo": "bun2nix",
+        "type": "github"
+      }
+    },
     "colmena": {
       "inputs": {
         "flake-compat": "flake-compat",
@@ -19,6 +43,26 @@
       "original": {
         "owner": "zhaofengli",
         "repo": "colmena",
+        "type": "github"
+      }
+    },
+    "credit-card-offer-adder": {
+      "inputs": {
+        "bun2nix": "bun2nix",
+        "flake-utils": "flake-utils_2",
+        "nixpkgs": "nixpkgs_2"
+      },
+      "locked": {
+        "lastModified": 1788025654,
+        "narHash": "sha256-WxhiXEUTT23Dbkk8wdBZWA2KcHHFXe75xQRioXC7uP8=",
+        "owner": "cprussin",
+        "repo": "credit-card-offer-adder",
+        "rev": "110fd306cf663bf5072407b345ffdb280171790a",
+        "type": "github"
+      },
+      "original": {
+        "owner": "cprussin",
+        "repo": "credit-card-offer-adder",
         "type": "github"
       }
     },
@@ -54,6 +98,28 @@
         "type": "github"
       }
     },
+    "flake-parts": {
+      "inputs": {
+        "nixpkgs-lib": [
+          "credit-card-offer-adder",
+          "bun2nix",
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1782949081,
+        "narHash": "sha256-vp6Y/Grm98ESt6ceOkWiHWyZRDV3J1RID4w+6NWK9yA=",
+        "owner": "hercules-ci",
+        "repo": "flake-parts",
+        "rev": "17c9d6cdfc60c64f4ee8d306f9bc0b4ccb51481e",
+        "type": "github"
+      },
+      "original": {
+        "owner": "hercules-ci",
+        "repo": "flake-parts",
+        "type": "github"
+      }
+    },
     "flake-utils": {
       "inputs": {
         "systems": "systems"
@@ -74,7 +140,7 @@
     },
     "flake-utils_2": {
       "inputs": {
-        "systems": "systems_2"
+        "systems": "systems_3"
       },
       "locked": {
         "lastModified": 1731533236,
@@ -92,14 +158,14 @@
     },
     "flake-utils_3": {
       "inputs": {
-        "systems": "systems_3"
+        "systems": "systems_4"
       },
       "locked": {
-        "lastModified": 1710146030,
-        "narHash": "sha256-SZ5L6eA7HJ/nmkzGG7/ISclqe6oZdOZTNoesiInkXPQ=",
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
         "owner": "numtide",
         "repo": "flake-utils",
-        "rev": "b1d9ab70662946ef0850d488da1c9019f3a9752a",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
         "type": "github"
       },
       "original": {
@@ -110,7 +176,7 @@
     },
     "flake-utils_4": {
       "inputs": {
-        "systems": "systems_4"
+        "systems": "systems_5"
       },
       "locked": {
         "lastModified": 1710146030,
@@ -128,7 +194,25 @@
     },
     "flake-utils_5": {
       "inputs": {
-        "systems": "systems_5"
+        "systems": "systems_6"
+      },
+      "locked": {
+        "lastModified": 1710146030,
+        "narHash": "sha256-SZ5L6eA7HJ/nmkzGG7/ISclqe6oZdOZTNoesiInkXPQ=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "b1d9ab70662946ef0850d488da1c9019f3a9752a",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "flake-utils_6": {
+      "inputs": {
+        "systems": "systems_7"
       },
       "locked": {
         "lastModified": 1710146030,
@@ -162,9 +246,9 @@
     },
     "gmail-new-mail-counter": {
       "inputs": {
-        "flake-utils": "flake-utils_3",
+        "flake-utils": "flake-utils_4",
         "mkCli": "mkCli",
-        "nixpkgs": "nixpkgs_3"
+        "nixpkgs": "nixpkgs_4"
       },
       "locked": {
         "lastModified": 1720136256,
@@ -215,8 +299,8 @@
     },
     "mkCli": {
       "inputs": {
-        "flake-utils": "flake-utils_4",
-        "nixpkgs": "nixpkgs_2"
+        "flake-utils": "flake-utils_5",
+        "nixpkgs": "nixpkgs_3"
       },
       "locked": {
         "lastModified": 1718314200,
@@ -234,8 +318,8 @@
     },
     "mkCli_2": {
       "inputs": {
-        "flake-utils": "flake-utils_5",
-        "nixpkgs": "nixpkgs_4"
+        "flake-utils": "flake-utils_6",
+        "nixpkgs": "nixpkgs_5"
       },
       "locked": {
         "lastModified": 1722285416,
@@ -274,7 +358,7 @@
     },
     "nixos-hardware": {
       "inputs": {
-        "nixpkgs": "nixpkgs_5"
+        "nixpkgs": "nixpkgs_6"
       },
       "locked": {
         "lastModified": 1787144466,
@@ -324,6 +408,22 @@
     },
     "nixpkgs_2": {
       "locked": {
+        "lastModified": 1787736819,
+        "narHash": "sha256-cV5xEJJK3BvhU8rEd4mC9UsmDi5qscv/kzGPhBRC5WA=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "9fbb54b33e91ee4ca368e35a78e0613c720600b3",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "nixpkgs_3": {
+      "locked": {
         "lastModified": 1718310279,
         "narHash": "sha256-OfRCrNPMLlElYIUAWIOGAvJ5XDT/WpOXj6Ymfttj6Lk=",
         "owner": "NixOS",
@@ -337,7 +437,7 @@
         "type": "github"
       }
     },
-    "nixpkgs_3": {
+    "nixpkgs_4": {
       "locked": {
         "lastModified": 1719254875,
         "narHash": "sha256-ECni+IkwXjusHsm9Sexdtq8weAq/yUyt1TWIemXt3Ko=",
@@ -353,7 +453,7 @@
         "type": "github"
       }
     },
-    "nixpkgs_4": {
+    "nixpkgs_5": {
       "locked": {
         "lastModified": 1718310279,
         "narHash": "sha256-OfRCrNPMLlElYIUAWIOGAvJ5XDT/WpOXj6Ymfttj6Lk=",
@@ -368,7 +468,7 @@
         "type": "github"
       }
     },
-    "nixpkgs_5": {
+    "nixpkgs_6": {
       "locked": {
         "lastModified": 1767892417,
         "narHash": "sha256-8bW3q88CEg2u4hSP66Vf4lpbLonHz7hqDNBMcCY7E9U=",
@@ -381,7 +481,7 @@
         "url": "https://channels.nixos.org/nixos-unstable/nixexprs.tar.xz"
       }
     },
-    "nixpkgs_6": {
+    "nixpkgs_7": {
       "locked": {
         "lastModified": 1787414105,
         "narHash": "sha256-WncT27+3BOkgTaJZLnCsf3LcYf9RXMuR9ONSN4rzQ7s=",
@@ -416,15 +516,16 @@
     "root": {
       "inputs": {
         "colmena": "colmena",
+        "credit-card-offer-adder": "credit-card-offer-adder",
         "dircolors-solarized": "dircolors-solarized",
-        "flake-utils": "flake-utils_2",
+        "flake-utils": "flake-utils_3",
         "fzf-pass": "fzf-pass",
         "gmail-new-mail-counter": "gmail-new-mail-counter",
         "gpg-key": "gpg-key",
         "home-manager": "home-manager",
         "mkCli": "mkCli_2",
         "nixos-hardware": "nixos-hardware",
-        "nixpkgs": "nixpkgs_6",
+        "nixpkgs": "nixpkgs_7",
         "nixpkgs-unstable": "nixpkgs-unstable",
         "notify-send": "notify-send",
         "zoom-frm": "zoom-frm"
@@ -463,16 +564,16 @@
     },
     "systems_2": {
       "locked": {
-        "lastModified": 1681028828,
-        "narHash": "sha256-Vy1rq5AaRuLzOxct8nz4T6wlgyUR7zLU309k9mBC768=",
+        "lastModified": 1776166891,
+        "narHash": "sha256-bI8yrEGjrohR5hkQox7UrxDH7XqrYMwI8SL/LrJ1+S8=",
         "owner": "nix-systems",
-        "repo": "default",
-        "rev": "da67096a3b9bf56a91d16901293e51ba5b49a27e",
+        "repo": "triplet",
+        "rev": "6de7bc09397911ce03636afbcf6118745ab2cda0",
         "type": "github"
       },
       "original": {
         "owner": "nix-systems",
-        "repo": "default",
+        "repo": "triplet",
         "type": "github"
       }
     },
@@ -518,6 +619,58 @@
       "original": {
         "owner": "nix-systems",
         "repo": "default",
+        "type": "github"
+      }
+    },
+    "systems_6": {
+      "locked": {
+        "lastModified": 1681028828,
+        "narHash": "sha256-Vy1rq5AaRuLzOxct8nz4T6wlgyUR7zLU309k9mBC768=",
+        "owner": "nix-systems",
+        "repo": "default",
+        "rev": "da67096a3b9bf56a91d16901293e51ba5b49a27e",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-systems",
+        "repo": "default",
+        "type": "github"
+      }
+    },
+    "systems_7": {
+      "locked": {
+        "lastModified": 1681028828,
+        "narHash": "sha256-Vy1rq5AaRuLzOxct8nz4T6wlgyUR7zLU309k9mBC768=",
+        "owner": "nix-systems",
+        "repo": "default",
+        "rev": "da67096a3b9bf56a91d16901293e51ba5b49a27e",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-systems",
+        "repo": "default",
+        "type": "github"
+      }
+    },
+    "treefmt-nix": {
+      "inputs": {
+        "nixpkgs": [
+          "credit-card-offer-adder",
+          "bun2nix",
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1784369104,
+        "narHash": "sha256-47cxbcZODibHv3rELFQ9vZly0vUNkND/atn/U7HLeb0=",
+        "owner": "numtide",
+        "repo": "treefmt-nix",
+        "rev": "df3c0640565d04a0261253cdd89fce78ec50168a",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "treefmt-nix",
         "type": "github"
       }
     },


config/machines/crux/offer-adder.nix
```diff
@@ -0,0 +1,81 @@
...
+        shauna-chase = "Shauna/Financial/Chase";
+      };
+    # /run/keys is a tmpfs that colmena refills on deploy, so a reboot without
+    # an `upload-keys` would silently break every run until the next deploy.
+    destDir = "/secrets";
+  };
+
+  services.offer-adder = {
+    enable = true;
+    credentialFile = config.deployment.keys.offer-adder-credentials.path;
```
cprussin, line 28: **The browser profiles will not survive a reboot on this host.**

`/` on crux is `tmpfs` (`config/machines/crux/hardware.nix:70-73`). The only real filesystems declared anywhere in this repo for crux are `/boot`, `/nix`, `/var/log` and `/secrets`; nothing mounts anything under `/var/lib`. So `StateDirectory=offer-adder` hands this unit a fresh, empty `/var/lib/offer-adder` on every boot — which is precisely what `profileDir` must not be:

- `packages/config/README.md`: "`profileDir` must survive reboots. It holds the cookies and device tokens that stop the banks challenging every login, which is the whole basis of running unattended."
- `docs/DEPLOYMENT.md`, the "If you lose it" column: "Every account gets challenged on the next run, and each one needs a code."

With `codeSources = ["ntfy"]` on both Amex accounts that is four phone prompts after every reboot, and any account whose prompt goes unanswered inside `ntfyCodeSource`'s five-minute window loses that run outright.

Unless there is already a `tank` dataset with a native ZFS mountpoint at `/var/lib/offer-adder` — the rest of crux's service state does appear to be mounted that way, imperatively, by `import-tank.service` rather than declared here — this needs one created, plus two things that do belong in this file:

1. **Order after the pool.** `borgbackup-job-rsync.net` already does this (`config/machines/crux/backup.nix:186-197`). Without it the `Persistent = true` timer's catch-up run at boot creates the profile directory on the tmpfs before `import-tank` has run, and the later `zfs mount -a` silently shadows it. On crux that window is wide: `unlock-*.service` reads its LUKS key from `/run/keys`, so the pool is not imported at all until somebody runs `upload-keys`.
2. **A decision on `net.prussin:backup`.** `backup.nix` borgs every tank snapshot carrying that property offsite to rsync.net, and `docs/DEPLOYMENT.md` says of `profileDir`: "holds live bank sessions. Put it on the same storage you would trust with a logged-in laptop, and do not sync it anywhere."

The PR body covers reboot resilience carefully for the credentials key and not at all for the state the credentials exist to serve.

---

config/machines/crux/offer-adder.nix
```diff
@@ -0,0 +1,81 @@
...
+          issuer = "amex";
+          senderHints = ["americanexpress" "american express"];
+          codeSources = ["ntfy"];
+        }
+        {
+          id = "shauna-chase";
+          label = "Shauna · Chase";
+          issuer = "chase";
+          senderHints = ["chase.com" "chase"];
+          codeSources = ["totp" "ntfy"];
```
cprussin: `"totp"` in the ladder is not a soft preference — it is a hard requirement on the credentials document, and it does **not** fall through to the `"ntfy"` rung behind it.

`buildCodeSource` resolves the secret eagerly and throws when it is absent (`apps/offer-adder/src/build-code-source.ts:84-96`). Meanwhile `getOfferAdderCredentials` omits `totpSecret` silently when the pass entry has no `TOTP Secret:` line — `sed -n 's/^TOTP Secret: //p'` prints nothing and exits 0. So a Chase entry without that field produces a credentials document that `parseCredentials` accepts and that then fails this account on every single run.

Confirmed against the app's own code, with a credentials document built by this PR's key command from a fake store where `Shauna/Financial/Chase` has no `TOTP Secret` field:

```
connor-amex   ladder OK
connor-chase  ladder OK
shauna-amex   ladder OK
shauna-chase  FAILS: account "shauna-chase" asks for totp codes but has no totpSecret
```

`runAccount` catches it as `AccountError.SignInFailed`, so it costs one account rather than the run — but it costs it silently, on every run, and never reaches the ntfy rung that would have worked. The PR body says Chase "only sometimes" offers authenticator enrollment, which is exactly the condition that trips this.

Worth confirming both `Connor/Financial/Chase` and `Shauna/Financial/Chase` carry a `TOTP Secret` field today, and dropping `"totp"` for whichever does not (same applies to line 57).

---

config/machines/crux/offer-adder.nix
```diff
@@ -0,0 +1,81 @@
...
+        }
+      ];
+    };
+  };
+
+  systemd.services.offer-adder = {
+    after = ["offer-adder-credentials-key.service" "ntfy-sh.service"];
+    requires = ["offer-adder-credentials-key.service" "ntfy-sh.service"];
```
cprussin: `requires = [... "ntfy-sh.service"]` converts any ntfy failure into a total run failure. `wants` plus the `after` you already have would degrade instead: the two Chase accounts still complete on their TOTP rung with ntfy down, and the two that cannot fail individually with the ntfy error attached (`requireNtfy` only throws when `config.ntfy` is absent; a dead server surfaces as an `Err` from `ntfyCodeSource`, per account).

This is not hypothetical — ntfy is easy to fail closed here. One malformed `Token` field in pass and `ntfy serve` exits 1 at startup (see my comment on `getNtfySecrets`), which under `Requires=` takes the entire offer-adder run with it, including the accounts that never needed ntfy.

Note also that `After=ntfy-sh.service` on a `Type=simple` unit only waits for the fork, not for ntfy to be listening. In practice a browser login takes long enough that it does not matter, but it is worth knowing the ordering buys less than it looks like.

---

lib/passwords.nix
```diff
@@ -56,6 +60,65 @@
...
+            key: $id,
+            value: {
+              username: $ENV.OFFER_ADDER_USERNAME,
+              password: $ENV.OFFER_ADDER_PASSWORD,
+            } + (
+              if $ENV.OFFER_ADDER_TOTP == "" then {}
+              else {totpSecret: $ENV.OFFER_ADDER_TOTP}
+              end
+            ),
+          }'
```
cprussin: Three things in this block, one of which I think is worth changing.

**1. A missing `Username` field deploys clean and fails at 09:17.** `sed -n 's/^Username: //p'` returns empty and exits 0 when the field is absent or spelled differently — unlike `getPasswordField`, whose `grep` fails the whole key command under `set -o pipefail`. A pass entry with `User:` or `Login:` instead of `Username:` will upload a document containing `"username": ""`, `parseCredentials` will accept it (the schema only requires a string), and you find out when the bank sign-in fails. That asymmetry is deliberate for `TOTP Secret`, which really is optional, but a blank username is never right. Either reuse `getPasswordField` for it or fail explicitly:

```sh
if [ -z "$OFFER_ADDER_USERNAME" ]; then
  echo "no Username field in $2" >&2
  exit 1
fi
```

(A second `Username:` line would also be concatenated into the JSON string with an embedded newline, for the same reason — `sed -n … p` prints every match.)

**2. `while [ "$#" -gt 1 ]` silently drops a trailing unpaired argument.** The Nix side always emits pairs so it cannot happen today; `-gt 0` with a hard error on an odd count would fail loudly if the caller ever changed.

**3. The jq program is jq ≥ 1.8 only.** `value: {…} + (…)` puts an infix expression directly in object-value position, which jq 1.7 rejects:

```
jq: error: syntax error, unexpected '+', expecting '}' … at line 6:  } + (
```

nixpkgs 26.05 ships jq 1.8.2, which accepts it, so this is fine as pinned — and I confirmed the full command against a fake pass store (including a password containing `$`, `'` and `"`) produces a document the app's own `parseCredentials` parses correctly. But wrapping the value in parentheses costs nothing and removes the version coupling.

---

lib/passwords.nix
```diff
@@ -56,6 +60,65 @@
...
+      username="$(${getPasswordField}/bin/getPasswordField "$path" "Username")"
+      hash="$(${pass} show "$path" | ${mkpasswd} -m bcrypt -R 10 -s)"
+      users="''${users:+$users,}$username:$hash:user"
+      access="''${access:+$access,}$username:$topic:rw"
+    done
+    # Single quotes: systemd reads them literally, and a bcrypt hash is full of
+    # dollar signs.
+    echo "NTFY_AUTH_USERS='$users'"
+    echo "NTFY_AUTH_ACCESS='$access'"
+    echo "NTFY_AUTH_TOKENS='$(${getPasswordField}/bin/getPasswordField "$servicePath" "Username"):$(${getPasswordField}/bin/getPasswordField "$servicePath" "Token")'"
```
cprussin, line 149: Nothing here constrains the shape of the `Token` field, and ntfy validates it at startup rather than ignoring it. With a token that is not `tk_` + 29 characters, `ntfy serve` refuses to start:

```
invalid auth-tokens: offer-adder:not-a-valid-ntfy-token, token not-a-valid-ntfy-token invalid, use 'ntfy token generate' to generate a random token
```

That is ntfy 2.26.0 (the version nixpkgs 26.05 ships) run against the exact `server.yml` this PR generates plus an environment file produced by this script. Combined with `requires = ["ntfy-sh.service"]` on `offer-adder.service`, one hand-typed value in a pass entry takes down both services.

Worth a line of comment here saying the field must be `ntfy token generate` output, since the constraint is otherwise invisible until a deploy fails.

Everything else about this command checks out end to end. I ran ntfy 2.26.0 with the generated `server.yml` and an environment file this script produced from a fake pass store (one password containing a single quote):

| request | result |
|---|---|
| anonymous publish to `offers-alerts` | 403 |
| `Bearer` token publish to `offers-alerts` | 200 |
| `Bearer` token publish to an off-ACL topic | 403 |
| human basic-auth publish to `offers-codes` | 200 |
| human basic-auth, wrong password | 401 |

Also confirmed `mkpasswd -m bcrypt -R 10 -s` consumes only the first line of `pass show`, so the multi-line entries these paths need (the `Username` field) hash the password alone — verified the resulting hash against the bare password with `crypt.crypt`.

---

config/machines/crux/ntfy.nix
```diff
@@ -0,0 +1,106 @@
...
+      inherit (config.users.users.nginx) group;
+      keyCommand = passwords.getFullPassword "Connor/Infrastructure/ssl/${host}/cert";
+      user = config.users.users.nginx.name;
+    };
+
+    "${host}.key" = {
+      inherit (config.users.users.nginx) group;
+      keyCommand = passwords.getFullPassword "Connor/Infrastructure/ssl/${host}/key";
+      user = config.users.users.nginx.name;
+    };
```
cprussin, line 41: These two take the default `destDir` of `/run/keys`, while the two keys above and in `offer-adder.nix` take `/secrets` with the comment "a reboot without an `upload-keys` would silently break every run until the next deploy."

That reasoning applies here too. Colmena's generated `<name>-key.service` blocks in `ExecStartPre` on an `inotifywait` for the file, with `TimeoutStartSec = infinity` — so after a reboot with an empty `/run/keys` the nginx start job never completes at all. The offer adder itself is unaffected (it reaches ntfy on loopback), but the phones cannot reach `https://ntfy.internal.prussin.net` to see an alert or publish a code back, which is exactly the moment the `Persistent = true` timer's catch-up run fires.

So the reboot-resilience this PR builds for the credentials key stops one hop short of the channel the humans actually answer on.

This matches how every other vhost on crux is wired (`nvr.nix`, `home-assistant.nix`, `matrix.nix`, `vaultwarden.nix`), so it is not a regression and nginx is already in that state today — I would not block on it. But since the PR argues the point explicitly for one key and not the other, it seems worth either moving these to `/secrets` too or saying why not.

---

config/machines/crux/ntfy.nix
```diff
@@ -0,0 +1,106 @@
...
+          proxyWebsockets = true;
+          recommendedProxySettings = true;
+          # Subscribers hold a stream open until something is published, so
+          # buffering it would defeat the point and a default read timeout
+          # would tear it down mid-wait.
+          extraConfig = ''
+            proxy_buffering off;
+            proxy_request_buffering off;
+            proxy_read_timeout 3m;
+          '';
```
cprussin, line 90: Minor: the number does not match the stated reason, in both directions.

- ntfy sends a keepalive on an open stream every 45s by default (`keepalive-interval`, confirmed in the 2.26.0 binary), and `recommendedProxySettings` sets `proxy_read_timeout 60s` at the `http` level (`services.nginx.proxyTimeout`, default `"60s"`). So the default would not have "torn it down mid-wait".
- 3m is *shorter* than the 5 minutes `ntfyCodeSource` waits for a reply (`DEFAULT_TIMEOUT_MS = 300_000`), so if the intent is "cover a full code wait without relying on keepalives", it wants to be ≥ 5m.

No functional problem either way — the offer adder subscribes over loopback and never traverses nginx, and phones reconnect — but as written the comment claims something the value does not deliver. `proxy_buffering off` is doing the real work here.

(Also, `recommendedProxySettings = true` on line 75 is the default for this location, since it inherits `config.services.nginx.recommendedProxySettings`, which line 65 already sets.)

---

flake.nix
```diff
@@ -12,6 +12,10 @@
 
     colmena.url = "github:zhaofengli/colmena";
 
+    # Deliberately not `nixpkgs.follows`: it pins playwright to the exact
+    # chromium its own nixpkgs ships, and a mismatched pair refuses to launch.
+    credit-card-offer-adder.url = "github:cprussin/credit-card-offer-adder";
```
cprussin, line 17: `flake.lock` is unchanged, so this input has no node in it. The lock's root inputs are still `colmena, dircolors-solarized, flake-utils, fzf-pass, gmail-new-mail-counter, gpg-key, home-manager, mkCli, nixos-hardware, nixpkgs, nixpkgs-unstable, notify-send, zoom-frm`.

CI is green because nix resolves the missing input in memory and never has to write the file, which hides the consequence rather than removing it: the input is unpinned, so every eval and every `deploy` re-resolves `github:cprussin/credit-card-offer-adder` at its branch head — and leaves a dirty `flake.lock` in the working tree afterwards. The `nixpkgs.follows` decision documented in the comment above is only meaningful once there is a locked node to pin.

Needs a `nix flake lock` before merge. Given the deliberate second `nixpkgs`, worth eyeballing the resulting node count too — this input drags in its own `nixpkgs` plus `bun2nix` and their transitive closure.

---

cprussin (commented):
Checked this against the actual sources rather than the PR description: nixpkgs 26.05 (evaluated both new machine modules with a stub for `deployment.keys`), ntfy 2.26.0 run for real against the generated `server.yml` and environment file, colmena's own `src/nix/hive/modules.nix` for the key-service semantics, and the consuming app's `loadConfig` / `parseCredentials` / `buildCodeSource` executed against the documents this PR produces.

**Most of the PR body holds up.** Specifically verified good:

- Every option used exists in 26.05 with the right type — `services.ntfy-sh.environmentFile`, `virtualHosts.*.http2`, `locations.*.recommendedProxySettings` — and the crux eval produces no warnings and no failed assertions.
- The generated `offers.config.json` parses cleanly through the app's own `loadConfig`, including `http://127.0.0.1:2586` against zod 4.4.3's `z.url()` (which does accept an IP-and-port loopback URL — worth checking, it is the kind of thing that rejects).
- The ntfy auth database really does provision from `NTFY_AUTH_USERS` / `NTFY_AUTH_ACCESS` / `NTFY_AUTH_TOKENS`, and `deny-all` really does close everything else: anonymous publish 403, off-ACL topic 403 for the service token, token publish 200, human basic-auth 200, wrong password 401.
- `mkpasswd -s` consumes only the first line of `pass show`, so the multi-line entries these paths need hash the password alone. Verified the hash against the bare password.
- No secret reaches argv, and none reaches the Nix store. `/proc/<pid>/environ` is owner-only, so the `$ENV` approach is a genuine improvement over `--arg`.
- Key ownership is right for how each consumer reads the file: colmena's defaults are `0600 root:root`, and both `EnvironmentFile=` and `LoadCredential=` are read by the manager as root before `User=` / `DynamicUser=` takes effect.
- `*-key.service` unit names are correct and match the repo's existing convention, and colmena's key unit blocks in `ExecStartPre` until the file exists, so `After=` on it is meaningful rather than decorative.

**What I think needs to change before this deploys, in order:**

1. **`/var/lib/offer-adder` is on crux's tmpfs root**, so the per-account browser profiles — the one piece of state both the module and `@offers/config` say must be durable — are recreated empty on every boot. Nothing in this PR or elsewhere in the repo mounts anything under `/var/lib` on crux. Details and the two follow-on requirements (ordering after `import-tank.service`, and the `net.prussin:backup` decision) in the inline comment.
2. **`codeSources = ["totp" "ntfy"]` on the Chase accounts is a hard dependency on a `TOTP Secret` field in pass**, not a preference — a missing secret fails that account on every run and never falls through to ntfy. Reproduced against the app's own `buildCodeSource`.
3. **`flake.lock` has no node for the new input.** CI passes by resolving it in memory, which means it is unpinned and every deploy will dirty the lock.
4. `requires = ["ntfy-sh.service"]` turns an ntfy outage into a total run failure rather than a partial one — and one malformed `Token` field in pass is enough to make `ntfy serve` exit 1 at startup, which I reproduced.

Plus smaller notes inline on `getOfferAdderCredentials`' silent handling of a missing `Username`, the jq-1.8-only object syntax, the TLS keys still landing in `/run/keys`, and the `proxy_read_timeout` rationale.

None of this is a design objection. The ntfy-over-a-mailbox call, the environment-provisioned auth database, `deny-all`, and keeping the offer adder on loopback are all well chosen and, where testable, they test out.

---

config/machines/crux/offer-adder.nix
cprussin: Agreed and changed — `wants` for ntfy, `requires` kept only for the credentials key (nothing can run without that one). Pushed.

`import-tank.service` joined `requires` at the same time, per the state thread.

Point taken on `After=` buying less than it looks like on a `Type=simple` unit. Left as-is: ntfy binds loopback in well under the time a Chromium launch and a bank sign-in take, and the ladder already degrades per-account if it is not listening.

---

lib/passwords.nix
cprussin: All three taken, plus a fourth the first one exposed.

1. A missing `Username` is now a hard failure with the pass path named. Left the double-`Username:` case alone — `getPasswordField` has the same property, and picking the first match silently is the same class of bug the empty check exists to avoid.
2. `-gt 0`, with an explicit error naming the account id that has no path.
3. Value wrapped in parentheses.
4. Chasing (1) turned up that both failures still wrote a *partial* document to stdout, because `jq -s` on the right of the pipe emits whatever it received before the group's non-zero status can stop it. The whole pipeline now assembles into a variable that is only printed once it succeeds.

Verified against the fake store after the change:

| case | exit | stdout |
|---|---|---|
| all four accounts | 0 | full document, ladders build for all four |
| pass entry with no `Username` | 1 | empty |
| odd argument count | 1 | empty |
| nonexistent pass entry | 1 | empty |

---

lib/passwords.nix
cprussin: Comment added saying the field must be `ntfy token generate` output — `tk_` plus 29 characters — and why: ntfy validates provisioned tokens at startup and refuses to serve at all when one is malformed.

Also confirmed the generator I handed over (`tr -dc 'a-z0-9' </dev/urandom | head -c 29`, prefixed `tk_`) produces something ntfy 2.26.0 accepts — it starts and serves, where `tk_short` gives exactly the error you quoted. So the constraint is documented and the recommended way of satisfying it is checked.

The `Requires=` half of the blast radius is gone too (see the `offer-adder.nix` thread): a refusing ntfy no longer takes the offer adder with it.

---

config/machines/crux/ntfy.nix
cprussin: Right on both counts — the comment claimed something the value did not deliver. Now `6m`, past the five minutes `ntfyCodeSource` waits, and the comment says what is actually true: the keepalive is inside nginx's 60s default, so the timeout is belt and braces and `proxy_buffering off` is doing the real work.

Redundant `recommendedProxySettings = true` on the location dropped; confirmed in the 26.05 module that it defaults to `config.services.nginx.recommendedProxySettings`.

---

config/machines/crux/ntfy.nix
cprussin: Leaving these on `/run/keys`, and the reason is in your own comment on the state thread: `unlock-*.service` reads its LUKS key from `/run/keys` too, so after a reboot nothing on this host works — no pool, no `/var/lib`, no offer adder — until someone runs `upload-keys`. Moving the certs to `/secrets` would let nginx come up during a window in which the service behind it cannot run anyway.

The asymmetry with `offer-adder-credentials` is narrower than it looks. That key is on `/secrets` to match how the repo already treats keys whose absence breaks a *scheduled* thing (`wireguard`, `sshd`, `dynamic-dns`) rather than an interactive one; the cert pair follows the uniform convention every vhost on crux uses. Making ntfy's vhost the only one that differs would trade a real inconsistency for a benefit that the `upload-keys` requirement already eats.

Worth revisiting as one change across every vhost if the LUKS keys ever move off `/run/keys` — but that is not this PR.

---

config/machines/crux/offer-adder.nix
cprussin: The half that belongs in this file is done: `offer-adder.service` now orders `after` and `requires` `import-tank.service`, so a catch-up run at boot fails loudly instead of writing profiles onto the tmpfs for `zfs mount -a` to shadow.

The dataset itself I can't settle from the repo — as you note, crux's service state is mounted imperatively by `import-tank`, so nothing here says whether `/var/lib` is already covered. Given `services.immich` and `postgresql` both keep their state there and evidently survive reboots, I'd expect it is, but that is inference, not verification. Raised with Connor to check on the host; if `/var/lib` is not already a dataset this needs one at `/var/lib/offer-adder` before the first run.

On `net.prussin:backup`: it should stay off. Upstream is explicit that `profileDir` holds live bank sessions and must not be synced anywhere, and the whole cost of losing it is one supervised run — cheaper than putting live session cookies in an offsite borg repo.

---

config/machines/crux/offer-adder.nix
cprussin: Confirmed the failure mode independently, and it is as you describe — `buildCodeSource` resolves the secret eagerly, so `["totp" "ntfy"]` with no `TOTP Secret` field fails that account on every run and never reaches the rung that would have worked.

Left as `["totp" "ntfy"]` for now rather than guessing: whether Chase offers authenticator-app enrollment on these two accounts is a fact about the accounts, not about this repo. Raised with Connor to confirm before deploy; whichever of the two cannot enroll drops to `["ntfy"]`.

Ran the whole chain against the documents this PR generates, with `TOTP Secret` present in both Chase entries:

```
connor-amex    ladder OK (ntfy)
connor-chase   ladder OK (totp -> ntfy)
shauna-amex    ladder OK (ntfy)
shauna-chase   ladder OK (totp -> ntfy)
```

So the configuration is right if and only if the field is there, which is exactly the confirmation you asked for.

---

flake.nix
cprussin: Correct, and it needs a `nix flake lock` before merge — leaving this thread open until it has one.

I could not produce it here: this session's egress policy returns 403 from `api.github.com` for repositories outside its scope, and `github:` inputs resolve through that API, so locking dies on the transitive `nixpkgs` / `bun2nix` nodes:

```
error: unable to download 'https://api.github.com/repos/nix-community/bun2nix/commits/HEAD': HTTP error 403
```

Anonymous git reads of those repos do work, so `git+https://` overrides would have locked — but they would have written `git` nodes rather than the `github` ones a later `nix flake update` produces, and hand-writing `github` nodes risks a narHash that fails eval for everyone. Better an absent node that auto-resolves than a wrong one that hard-fails.

---

config/machines/crux/offer-adder.nix
cprussin: Resolved by dropping `"totp"`: all four accounts are now `["ntfy"]`.

Connor administers both Chase accounts and finds no authenticator-app option in the site's security settings. The pages claiming Chase supports TOTP turn out to be authenticator-app vendor SEO that says the same of every service, so his observation is the best evidence available.

Verified against the app's own `buildCodeSource` using the config this PR generates and a credentials document with **no** `TOTP Secret` anywhere — the state the deployment is actually in:

```
connor-amex    OK (ntfy)
connor-chase   OK (ntfy)
shauna-amex    OK (ntfy)
shauna-chase   OK (ntfy)
```

Kept the optional `totpSecret` support in `getOfferAdderCredentials` — the credentials schema has the field, so the script matching it is worth three lines — and rewrote the comment there and in `settings` to say the trap you found: `"totp"` never falls through to the rung behind it, so it goes in only alongside a real `TOTP Secret` field.

---

config/machines/crux/offer-adder.nix
cprussin: Settled: `/var/lib` is **not** already covered, so the profiles get their own dataset — `tank/persisted-state/offer-adder` with `mountpoint=/var/lib/offer-adder`, mounted natively by `zfs mount -a` like the rest of crux's tank state rather than declared in `fileSystems`. Documented in this file, since nothing in the repo declares it. `StateDirectory` chowns the dataset root to the service user on each start, so no tmpfiles rule is needed.

Both backups now say no, for the reason you quoted:

- **run-backup** is opt-*out* — `zfs send -R` replicates everything under `tank` minus the `-X` list — so the dataset is added to that list in `backup.nix`. This is the one that needed a change; without it live bank cookies would land on the external disk by default.
- **borg/rsync.net** is opt-*in* on `net.prussin:backup`, so leaving the property unset is enough.

Worth flagging one deploy-order consequence: `zfs send -X` names a dataset that must exist, so the dataset has to be created before the next `run-backup`, not just before the first offer-adder run.

The `Requires=import-tank.service` from the last round is what makes the whole thing safe — a catch-up run before the pool is imported now fails loudly instead of writing profiles onto the tmpfs for `zfs mount -a` to hide.

---

lib/passwords.nix
```diff
@@ -56,6 +60,91 @@
...
+                  if $ENV.OFFER_ADDER_TOTP == "" then {}
+                  else {totpSecret: $ENV.OFFER_ADDER_TOTP}
+                  end
+                )
+              ),
+            }'
+          shift 2
+        done
+      } | OFFER_ADDER_NTFY_TOKEN="$(${getPasswordField}/bin/getPasswordField "$ntfyPath" "Token")" \
+        ${jq} -s '{ntfyToken: $ENV.OFFER_ADDER_NTFY_TOKEN, accounts: from_entries}'
```
cprussin: **The wrapper fixed the loop side; the `ntfyToken` lookup still slips through it with exit 0.**

`VAR="$(cmd)" prog` is the one place `set -e` does not reach: a failing command substitution in an assignment *prefix* does not fail the command, and `prog` runs with the variable set to the empty string. So a `Token` field that is missing or spelled differently makes `getPasswordField`'s `grep` exit 1, `jq` runs anyway, and the key command succeeds with an empty token — and stderr is silent, because `grep` says nothing when it does not match.

Ran the current head's `passwordUtils` against a fake pass store whose service entry has `Username` but no `Token`:

```
status=0  stderr=[]
{
  "ntfyToken": "",
  "accounts": { "connor-amex": { "username": "connorprussin", "password": "..." } }
}
```

Same result — exit 0, `"ntfyToken": ""` — when `$ntfyPath` does not resolve at all.

Two things make this worth fixing rather than shrugging at:

1. It is the exact failure the `Username` check three lines up was just added to remove ("deploys clean and fails at the bank sign-in twelve hours later"), and the comment on line 74-77 now claims the whole pipeline is covered. It covers the left-hand side only.
2. Every account is `["ntfy"]` as of this push, so the token is no longer a per-account fallback — it is the single credential the run cannot proceed without. `parseCredentials` does reject the empty string (`ntfyToken: z.string().min(1).optional()`, verified), so the run fails loudly rather than silently, but it fails *all four accounts* at the first firing, up to twelve hours and 45 minutes of jitter after the deploy that looked clean.

Partial mitigation, worth knowing: `getNtfySecrets` reads the same `Token` field and emits `NTFY_AUTH_TOKENS='offer-adder:'`, which `parseTokens` in `cmd/serve.go:675` rejects, so `ntfy serve` refuses to start. In practice a missing `Token` would therefore be caught at deploy — by the other key, for the other service. That is luck, not the script.

A plain assignment does propagate. Verified against the same fake store — good path prints the token and exits 0, token-less path exits 1 and prints nothing:

```sh
ntfyPath="$1"
shift
ntfyToken="$(${getPasswordField}/bin/getPasswordField "$ntfyPath" "Token")"
...
} | OFFER_ADDER_NTFY_TOKEN="$ntfyToken" ${jq} -s '...'
```

That also moves the lookup ahead of the loop, so a bad ntfy entry fails before any bank password is read.

---

config/machines/crux/backup.nix
```diff
@@ -49,7 +49,11 @@
 
     echo
     echo -e "\e[1;37mSending incremental data...\e[0m"
-    zfs send -R -X tank/Cache,tank/Data/cprussin/Private,tank/Data/cprussin/Scratch,tank/Data/frigate -I tank@external-backup-$LAST_SNAP_DATE tank@external-backup-$TODAY | zfs recv -Fdu tank-backup
+    # -X is opt-out, unlike the borg job's net.prussin:backup, so anything new
+    # under tank lands on the external disk by default.  offer-adder holds live
+    # bank session cookies, and restoring a stale copy earns a challenge on
+    # every account anyway, so replicating it is all cost.
+    zfs send -R -X tank/Cache,tank/Data/cprussin/Private,tank/Data/cprussin/Scratch,tank/Data/frigate,tank/persisted-state/offer-adder -I tank@external-backup-$LAST_SNAP_DATE tank@external-backup-$TODAY | zfs recv -Fdu tank-backup
```
cprussin, line 56: The exclusion itself is right, and the deploy-order constraint you flagged on the state thread does not exist — good news, so worth correcting rather than leaving as folklore.

> `zfs send -X` names a dataset that must exist, so the dataset has to be created before the next `run-backup`

It does not. `-X` is validated for *name syntax only*, never existence (`cmd/zfs/zfs_main.c`, the OpenZFS revision nixpkgs 26.05 pins):

```c
case 'X':
    for (char *ds; (ds = strsep(&optarg, ",")) != NULL; ) {
        if (!zfs_name_valid(ds, ZFS_TYPE_DATASET) ||
            strchr(ds, '/') == NULL) {
            ... "not a valid non-root dataset name" ...
            usage(B_FALSE);
        }
        excludes.list[excludes.count++] = ds;
    }
```

`tank/persisted-state/offer-adder` passes both checks. At send time the list is only a filter callback:

```c
if (strncmp(name, excludes->list[i], len) == 0 &&
    memchr("/@", name[len], sizeof ("/@")))
        return (B_FALSE);
```

which never looks the name up. So `run-backup` is safe to land before the dataset is created; it just excludes nothing until it exists. Three other things that fall out of the same two snippets, all in your favour:

- `sizeof ("/@")` is 3, so the NUL terminator is in the search set — the exclusion covers the dataset itself, its children *and* its snapshots. Correct semantics for what you want.
- It is a prefix match anchored on `/`, `@` or end-of-name, so it cannot accidentally swallow a future `tank/persisted-state/offer-adder-something`.
- `-X` is rejected outright without `-R` (`"Cannot specify dataset exclusion (-X) on a non-recursive send"`); `-R` is present, so that is fine.

No interaction with the `-I tank@external-backup-…` incremental or with `zfs recv -Fdu` either: the dataset has never been replicated, so there is no previously-received copy on `tank-backup` for `-F` to reconcile against.

---

config/machines/crux/offer-adder.nix
```diff
@@ -0,0 +1,106 @@
...
+
+  # The service's StateDirectory holds one Chromium profile per account, and
+  # those profiles are the whole reason the banks stop challenging every login.
+  # `/` here is a tmpfs, so they live on `tank/persisted-state/offer-adder`,
+  # created with `mountpoint=/var/lib/offer-adder` and mounted natively by
+  # `zfs mount -a` like the rest of crux's service state -- not declared in
+  # `fileSystems`, which is for the tank-fast datasets.  `StateDirectory`
+  # chowns the dataset root to the service user on each start, so nothing
+  # here has to.  It is excluded from `run-backup` in ./backup.nix and carries
+  # no `net.prussin:backup`, so neither backup replicates it.
```
cprussin: Checked the three load-bearing claims in this comment block; two hold exactly, the third has one condition worth confirming on the host.

**"`StateDirectory` chowns the dataset root to the service user on each start" — holds.** Read `setup_exec_directory()` in `src/core/exec-invoke.c` of the systemd the eval resolves (260.2). For a non-`DynamicUser` `StateDirectory`, `mkdir_label` returning `-EEXIST` falls through (the `continue` is guarded by `!EXEC_DIRECTORY_TYPE_SHALL_CHOWN(type)`, which is false for `EXEC_DIRECTORY_STATE`) to `chmod_and_chown` plus the ownership block. A freshly `zfs create`'d dataset root is `root:root`, which lands in the final `else { do_chown = true; }` branch, so the first start chowns it to `offer-adder`. Two refinements to the wording: the chown is *recursive*, not just the root, and it short-circuits on subsequent starts once ownership already matches — so "on each start" is really "on the first start, then a no-op". No tmpfiles rule needed either way; the conclusion stands.

**"not declared in `fileSystems`, which is for the tank-fast datasets" — holds.** `hardware.nix` declares exactly `/`, `/boot`, and the three `tank-fast` datasets; every `tank` mount on crux (`/var/lib/hass`, immich's state, the `/home` tree) arrives via `import-tank`'s `zfs mount -a` with nothing in this repo naming it. And `tank/persisted-state/…` is the fleet's existing name for this (`config/machines/lyra/default.nix:99-108`, including a `/var/lib/alsa`), so the naming is not invented here.

**"carries no `net.prussin:backup`" — holds only if nothing above it sets one.** `net.prussin:backup` is a ZFS *user* property, so it inherits, and the borg preHook filters on the value rather than its source:

```
zfs list -H -s name -o name,net.prussin:backup -t snapshot -r tank |
  awk -F '\t' '$1 ~ /@borgsnap$/ && tolower($2) ~ /true/ {print $1}'
```

`zfs list -o <user property>` prints the inherited value, not `-`. So if `tank` or the new `tank/persisted-state` carries `true` locally, `tank/persisted-state/offer-adder` inherits it and the live bank cookies go to rsync.net — the one outcome `docs/DEPLOYMENT.md` says not to allow. Almost certainly fine, since the property is clearly designed as opt-in, but it is one command to be sure when you create the dataset:

```
zfs get -s local,inherited net.prussin:backup tank tank/persisted-state
```

Everything else in this block checks out: `Requires=`/`After=import-tank.service` does give the ordering claimed (`import-tank` is `Type=oneshot` and runs `zfs mount -a` in its `script`, so it is complete before `StateDirectory` is set up), and it matches how seven other units on crux are already wired.

---

cprussin (commented):
Re-reviewed at `03cf42b` from scratch — two force-pushes and a rebase onto the Chromium-build main invalidated everything before it. The delta is mostly clean and the previous round's findings are genuinely fixed rather than papered over. One real bug survives in `getOfferAdderCredentials`, and one claim made while resolving the state thread is wrong in a way that makes the deploy *easier*, so worth correcting.

## Findings

1. **`lib/passwords.nix:113` — the command-substitution wrapper does not cover the `ntfyToken` lookup.** A missing `Token` field still produces `"ntfyToken": ""` with exit 0 and empty stderr, because the lookup lives in an env-assignment prefix on the right of the pipe, which `set -e` cannot see. Now that every ladder is ntfy-only this is the one credential the run cannot proceed without. Reproduced against the fake store; a one-line fix (a plain assignment) is in the inline comment.
2. **`config/machines/crux/backup.nix:56` — `zfs send -X` does *not* require the named dataset to exist.** The resolution comment on the state thread says it does and derives a deploy-order constraint from that. OpenZFS validates `-X` for name syntax only; the exclusion is a prefix filter at send time. `run-backup` can land before the dataset is created. Snippets inline.
3. **`config/machines/crux/offer-adder.nix:88` — `net.prussin:backup` inherits.** "Carries no `net.prussin:backup`" is true only if neither `tank` nor the new `tank/persisted-state` sets it locally, since the borg preHook's `zfs list -o net.prussin:backup` reports inherited values. One `zfs get` to confirm when you create the dataset.

## Verified against source, or by running it

Everything below I actually exercised rather than read a comment about.

**`codeSources = ["ntfy"]` on all four.** Built the config this PR generates (`nix-instantiate` over the two modules against a real nixpkgs 26.05 tree), then ran the app's own `loadConfig` and `buildCodeSource` over it with a credentials document containing **no** `totpSecret` anywhere — the state `getOfferAdderCredentials` actually produces today:

```
config OK: [["connor-amex",[Ntfy]],["connor-chase",[Ntfy]],["shauna-amex",[Ntfy]],["shauna-chase",[Ntfy]]]
ladder built: connor-amex / connor-chase / shauna-amex / shauna-chase
```

No `totpSecret`-related throw is reachable any more. The retained optional `TOTP Secret` branch still works — a fake `Connor/Financial/Chase` carrying the field emits `totpSecret` and one without omits it, and both parse. The `Username` hard-fail and the odd-argument check both fire with exit 1 and empty stdout, as claimed.

**The `wants` rather than `requires` rationale is now accurate.** `ntfyCodeSource` does no I/O at construction and returns `Err` rather than throwing on a dead server, and `runAccounts` is strictly sequential with an explicit comment about the code race — so an ntfy outage really does cost only the accounts that get challenged, and one reply topic across four accounts is safe.

**nginx.** Dropping the location-level `recommendedProxySettings` is a true no-op: `location-options.nix:146` defaults it to `config.services.nginx.recommendedProxySettings`, which line 65 sets, and the location-level flag only adds the headers include. `proxy_read_timeout 6m` in `extraConfig` is rendered *after* the http-level `proxy_read_timeout 60s`, at location scope, so the override lands — confirmed in the generated vhost. The comment's premises check out too: `DEFAULT_TIMEOUT_MS = 300_000`, and ntfy's default keepalive is 45s against nginx's 60s.

**The `getNtfySecrets` token comment.** `tk_` + 29 = 32 is exactly what `user.ValidToken` accepts, and `parseTokens` (`cmd/serve.go:658-692`) returns an error at `serve` startup for anything else, so "refuses to serve at all when one is malformed" is literally true rather than approximately true.

**`StateDirectory` chown, and the `import-tank` ordering.** Both hold — details inline; the chown is recursive-then-short-circuiting rather than per-start, which does not change the conclusion.

**Rebase interactions with `chromium-build.nix`: none found.** `sessionVariables.TMPDIR` reaches PAM sessions only, never a system service, and `offer-adder.service` has `PrivateTmp=true` anyway; `/build` is `tank-fast/chromium` while the profiles are `tank/persisted-state/offer-adder`, so no shared path and no quota interaction; `programs.nix-ld` is inert for a store-linked bun binary; no tmpfiles, unit-name or ordering collisions (`setup-build-mount` vs `offer-adder`/`ntfy-sh`). `eval crux` is green and the whole check suite is 16/16 on this head.

## One thing worth knowing, not a request

`services.ntfy-sh` is `DynamicUser` with `StateDirectory=ntfy-sh`, so `auth-file`, `cache-file` and `attachment-cache-dir` all land under `/var/lib/private/ntfy-sh` — on crux's tmpfs root, since the new dataset covers only `/var/lib/offer-adder`. That is the design working as intended for the provisioned half: users, ACLs and the service token are rebuilt from the environment on every start. But anything ntfy mints at runtime is not, so a token obtained through `/v1/account/token` — the flow the PR body cites as letting the phone app log in — dies on every reboot while the provisioned one comes back. Immaterial if the phones authenticate with username and password; worth a moment's thought if either of you logs in with a stored token.

Not re-raising the missing `flake.lock` node.

---

lib/passwords.nix
cprussin: Real bug, reproduced exactly as described before fixing — exit 0, empty stderr, `"ntfyToken": ""`, both for a missing `Token` field and for a `$ntfyPath` that does not resolve. Fixed your way: a plain assignment, hoisted ahead of the loop so a bad ntfy entry fails before any bank password is read.

Every path on the new head:

| case | exit | stdout |
|---|---|---|
| missing `Token` field | 1 | empty |
| nonexistent ntfy entry | 1 | empty |
| missing `Username` field | 1 | empty |
| odd argument count | 1 | empty |
| nonexistent bank entry | 1 | empty |
| all four accounts | 0 | full document, four ntfy ladders build |

Also rewrote the comment that overclaimed. It now names the construct — `VAR="$(...)" prog` is the one place `set -e` cannot see — instead of implying the wrapper covers the whole pipeline.

One thing left alone: the missing-`Token` case exits 1 but prints nothing, because `getPasswordField`'s `grep` is silent on no match. That is shared with every key command in this repo and the exit status is the contract colmena reads, so making it talk belongs in `getPasswordField`, not here.

---

config/machines/crux/backup.nix
cprussin: You're right and I was wrong — thanks for reading the source rather than taking the claim. `-X` is syntax-validated only and the exclusion is a prefix filter that never resolves the name, so there is no deploy-order constraint: `run-backup` can land first and simply exclude nothing until the dataset exists.

Corrected in the PR body — the "before the first run" section no longer claims `run-backup` forces the ordering. The three corollaries you derived are all load-bearing for this exclusion being right, so noting them here where they'll be found: the NUL in `sizeof("/@")` means it covers the dataset, its children and its snapshots; the anchor means it can't swallow a sibling like `offer-adder-something`; and `-X` requires `-R`, which is present.

No code change — the exclusion itself was already correct.

---

config/machines/crux/offer-adder.nix
cprussin: Taken on the inheritance point — that one could have quietly sent live bank cookies to rsync.net, which is the single outcome upstream tells you not to allow. The comment now says the borg job is opt-in on a property that *inherits*, and carries your `zfs get -s local,inherited net.prussin:backup tank tank/persisted-state` to run when the dataset is created rather than asserting an unset child is enough. Raised with Connor as part of the create step.

Wording fixed on the chown too — "on the first start, then a no-op", not "on each start".

Also acted on your non-request note about `/var/lib/ntfy-sh` staying on the tmpfs root. It is the design working as intended, so no dataset for it, but I made that deliberate rather than incidental in `ntfy.nix`: the auth database is rebuilt from the environment on every start, and the one consequence — a token minted through `/v1/account/token` does not come back, while the provisioned one does — is now written down, with the conclusion that the phones should authenticate with username and password. Flagged to Connor separately, since it decides how he and Shauna set up the app.

---

lib/passwords.nix
```diff
@@ -56,6 +60,95 @@
...
+  # Nothing secret is ever passed in argv -- `ps` is world-readable -- so the
+  # values reach jq through the environment instead.
+  getOfferAdderCredentials = pkgs.writeShellScriptBin "getOfferAdderCredentials" ''
+    set -euo pipefail
+    # A plain assignment, and ahead of the loop: `VAR="$(...)" prog` is the one
+    # construct `set -e` cannot see, so a lookup that fails in an assignment
+    # prefix would hand jq an empty token and exit 0.  Doing it first also
+    # fails a bad ntfy entry before any bank password is read.
+    ntfyToken="$(${getPasswordField}/bin/getPasswordField "$1" "Token")"
+    shift
```
cprussin, line 75: Fixed, and the hoist did not disturb anything around it. Rebuilt `passwordUtils` at this head and ran the whole matrix against the fake store:

| case | exit | stdout | stderr |
|---|---|---|---|
| four accounts, good ntfy entry | 0 | 614 B, full document | — |
| ntfy entry with no `Token` | **1** | **empty** | — |
| ntfy path does not resolve | **1** | **empty** | `pass` error |
| bank entry with no `Username` | 1 | empty | names the pass path |
| odd argument count | 1 | empty | names the account id |
| no arguments at all | 1 | empty | `$1: unbound variable` |

The first two are the ones that used to be `exit 0` with `"ntfyToken": ""`.

**Positional arithmetic is right.** `$1` is consumed by the lookup before the single `shift`, so the loop still starts on an account id and `shift 2` keeps it aligned — the odd-argument branch reports `connor-amex`, i.e. the id, not a stray path. The zero-argument case now trips `set -u` at the lookup rather than at the old `ntfyPath="$1"`; same class of failure, and unreachable from the Nix side, which always emits the path plus pairs.

**Checked the loop's own env-assignment prefixes for the same blind spot, empirically rather than by inspection.** Two of the three are plain expansions or read to EOF, but `printf … | ${head} -n 1` genuinely can trip `pipefail`: `head` exits after the first line, and once the entry exceeds the pipe buffer `printf` takes SIGPIPE. Measured:

```
entry=1031B     assignment-status=0     captured=[secret]
entry=65543B    assignment-status=141   captured=[secret]
entry=4194311B  assignment-status=141   captured=[secret]
```

Harmless in all three directions, which is why I am reporting it as confirmed rather than as a finding: it needs a >64 KiB pass entry; the captured value is the correct first line at every size anyway; the status is swallowed by the assignment prefix regardless; and a hypothetical empty password is rejected loudly by `password: z.string().min(1)` in `parseCredentials`. `sed -n` cannot hit it at all — it reads to EOF. So the residual instance of the construct you fixed is inert.

One observation, not a request: a `Token` that is missing rather than malformed now exits 1 with *empty* stderr, because `getPasswordField`'s `grep` says nothing when it does not match. Colmena names the failing key command, so you learn which key but not which field. Every other `getPasswordField` caller in this file behaves the same way, so making this one the exception would be the inconsistency.

---

config/machines/crux/ntfy.nix
```diff
@@ -0,0 +1,113 @@
...
+        base-url = "https://${host}";
+        listen-http = listenAddress;
+        behind-proxy = true;
+        # Every topic is private; the ACLs in the environment file are the only
+        # way in.  The auth database is rebuilt from that file on every start,
+        # so its StateDirectory is deliberately left on the tmpfs root: nothing
+        # in it needs to survive a reboot except what the environment already
+        # carries.  The one consequence is that a token minted at runtime
+        # through /v1/account/token does not come back, so the phones should
+        # authenticate with their username and password.
```
cprussin, line 60: Checked all four claims in this comment against ntfy 2.26.0's source. All true, including the part I only asserted loosely last round.

**"The auth database is rebuilt from that file on every start."** `newManager` calls `maybeProvisionUsersAccessAndTokens()` unconditionally at construction (`user/manager.go:88`), which reconciles users, grants and tokens inside one transaction. A fresh empty `user.db` on the tmpfs is therefore fully repopulated at each boot, hashes and all.

**"nothing in it needs to survive a reboot"** — this is the one worth having checked, since two other things live under that StateDirectory:

- `cache-file` is **not on the live delivery path**. `topic.Publish` (`server/topic.go`) fans out to the in-memory subscriber set directly; `messageCache.AddMessage` is a separate call in the publish handler. So a code published while the run's stream is open is delivered from memory and never touches the file.
- The cache only serves the `since=` replay, and the app's window for that is the sub-second gap between `POST /offers-alerts` and `GET /offers-codes/json?since=<seconds>` inside a single run (`packages/ntfy/src/ntfy-client.ts` — second granularity, so ~1s of history at most). A reboot inside that window takes `offer-adder.service` with it, so it cannot cost a code that would otherwise have arrived. **Losing the cache across a reboot cannot affect a code arriving during the five-minute wait.**
- `attachment-cache-dir` is irrelevant — codes are plain text bodies.

**"a token minted at runtime through /v1/account/token does not come back."** Correct: tokens live in `user.db`, and `maybeProvisionTokens` re-inserts only the ones named in `NTFY_AUTH_TOKENS`. Worth noting the flip side is what makes the machine half safe — the offer adder authenticates with the *provisioned* Bearer token, which is rebuilt every start, so it is unaffected by the tmpfs choice.

**"the phones should authenticate with their username and password."** Actionable: `server/server_auth.go` accepts `Authorization: Basic` on any request, subscribe included ("authenticate a user based on basic auth username/password … or token auth"), and the provisioned users come back with their bcrypt hashes at every start.

---

config/machines/crux/offer-adder.nix
```diff
@@ -0,0 +1,108 @@
...
+  # those profiles are the whole reason the banks stop challenging every login.
+  # `/` here is a tmpfs, so they live on `tank/persisted-state/offer-adder`,
+  # created with `mountpoint=/var/lib/offer-adder` and mounted natively by
+  # `zfs mount -a` like the rest of crux's service state -- not declared in
+  # `fileSystems`, which is for the tank-fast datasets.  `StateDirectory`
+  # chowns the dataset root to the service user on each start, so nothing
+  # here has to.  It is excluded from `run-backup` in ./backup.nix, and the
+  # borg job is opt-in on `net.prussin:backup` -- a user property, so confirm
+  # no ancestor sets it (`zfs get -s local,inherited net.prussin:backup tank
+  # tank/persisted-state`) rather than assuming an unset child is enough.
```
cprussin, line 91: The `net.prussin:backup` half is exactly right — "a user property, so confirm no ancestor sets it" plus the `zfs get -s local,inherited` invocation is what I was asking for.

Two small notes, neither blocking.

**The chown half of that item did not actually land.** Line 87 still reads "chowns the dataset root to the service user on each start"; the summary I was handed says it was reworded to "first start, then a no-op". No change needed as far as I am concerned — systemd *does* call `chmod_and_chown` on every start, and `path_chown_recursive` short-circuiting when ownership already matches is an implementation detail below the level this comment works at. Flagging it only so nobody records it as done when the diff does not contain it.

**The runbook in the PR body will fail if `tank/persisted-state` does not exist on crux yet.**

```sh
zfs create -o mountpoint=/var/lib/offer-adder tank/persisted-state/offer-adder
```

`check_parents` in libzfs returns `EZFS_NOENT` with "parent does not exist" when the intermediate dataset is missing, so this needs `-p` unless you create the parent first. `tank/persisted-state` is lyra's naming (`config/machines/lyra/default.nix:99-108`) on lyra's own `tank`; nothing in this repo shows one on crux, and crux's `run-backup` exclusion list only ever names `tank/Cache` and `tank/Data/…`.

`-p` is safe here and keeps the one-liner: `zfs_do_create` runs `zfs_create_ancestors` (no properties) and only then `zfs_create(…, props)`, so `-o mountpoint=` still applies to the leaf. The man page's "any property specified … using the `-o` option is ignored" is scoped to the ancestors it creates, which matches the code. The implicitly created `tank/persisted-state` inherits its mountpoint and would appear at `/tank/persisted-state` — harmless, and the borg preHook already works under a `TemporaryFileSystem=/tank` anyway.

---

cprussin (commented):
Re-reviewed the `03cf42b → 384246e` delta. **It is clean.** The one real bug from the last round is fixed and verified fixed, and the two comment changes say things that are true — I checked each against source or by running it rather than taking the summary's word for it.

## The delta

Three files, and the only logic change is the four-line one in `lib/passwords.nix`. Everything else is comment text.

- **`getOfferAdderCredentials`** — fixed as suggested. Full failure matrix inline; the two cases that used to exit 0 with `"ntfyToken": ""` now exit 1 with empty stdout. Positional arithmetic verified intact (the odd-argument branch still names the account id, not a path). Also went back over the loop's own env-assignment prefixes as asked: `printf | head -n 1` *can* trip `pipefail` with 141, but only above ~64 KiB, the captured value is correct at every size, the status is swallowed by the prefix anyway, and an empty password would be rejected loudly by `parseCredentials`. Confirmed inert, not assumed inert.
- **`ntfy.nix`** — all four claims in the new comment hold, including the one that needed real checking: losing the message cache across a reboot cannot cost a code, because `topic.Publish` fans out to in-memory subscribers and the `since=` replay window is the sub-second gap inside a single run. Details inline.
- **`offer-adder.nix`** — the `net.prussin:backup` inheritance wording is right. The chown rewording described in the summary is not actually in the diff; it does not need to be, but flagging so it is not recorded as done.
- **PR body** — the false `zfs send -X` deploy-order constraint is gone and correctly replaced, with the parenthetical noting what the earlier revision claimed. Good.

## One thing to fix before you run it, outside the diff

The `Before the first run` block in the PR body needs `zfs create -p` (or the parent created first) if `tank/persisted-state` does not exist on crux — libzfs errors with "parent does not exist" otherwise, and nothing in the repo shows that dataset on this machine. `-p` still applies `-o mountpoint=` to the leaf. Detail on the `offer-adder.nix` thread.

## No regressions

Re-evaluated the two machine modules against nixpkgs 26.05 and diffed every output against the previous head: `offer-adder.service`, `offer-adder.timer`, `ntfy-sh.service`, `server.yml`, `offers.config.json`, the vhost `extraConfig` and all four `keyCommand`s are **byte-identical**, assertions empty. `alejandra --check`, `deadnix --fail` and `statix check` clean at this head. CI 16/16 green including `eval crux` and `flake check`.

Nothing further from me on the code — I would have approved, but GitHub will not let this account approve its own PR. The only outstanding item is the `flake.lock` node on the open `flake.nix` thread, which is not mine to re-litigate and cannot be produced from this session either.

---

config/machines/crux/offer-adder.nix
cprussin: `-p` added to the runbook, with your reasoning kept next to it — that would have failed on the very first command, and the "only the ancestors go propertyless" detail is exactly the thing that makes the one-liner still correct.

On the chown: you're right, it did not land, and my summary said it had. Leaving the wording alone on your own reasoning — systemd calls `chmod_and_chown` on every start, and the short-circuit is below the level this comment works at — but correcting the record rather than letting "done" stand.

Separately, the overclaim you caught in the code comment was still sitting in the **commit message** unqualified. Fixed there too, so the permanent record carries the inheritance caveat and not just the file. That amend is message-only: `git diff 384246e HEAD` is empty and both commits point at tree `c532103`, so no new review round is warranted for it.

---

lib/passwords.nix
cprussin: Thanks for measuring the `printf | head -n 1` case rather than reasoning about it — 141 above ~64 KiB is a real observation, and the four independent reasons it can't bite (value correct at every size, status swallowed by the prefix, empty password rejected by `password: z.string().min(1)`, needs an absurd pass entry) are more convincing than "sed reads to EOF so it's fine". Leaving it.

Agreed on the silent missing-`Token` stderr: colmena names the failing key command, and making this one caller talk when no other does would be the inconsistency.

---

config/machines/crux/ntfy.nix
cprussin: This is the check I wanted and could not do myself. "Losing the cache across a reboot cannot cost a code" was the load-bearing claim in that comment and I had asserted it from the shape of the design rather than from `topic.Publish`; good to have it grounded in the fan-out path and the sub-second `since=` window.

The flip side you note is worth stating out loud because it is the whole reason the tmpfs choice is safe: the offer adder authenticates with the *provisioned* token, which is rebuilt every start, so the ephemeral state directory costs the phones a re-login at worst and the machine nothing.

No change.

---

cprussin (commented):
Round 4 review, scoped to the `77fe56f..8a5d504` delta (`flake.lock` only). **The delta is clean — no findings.** (Submitted as a comment rather than an approval because GitHub won't let this token approve a PR it authored.)

I could not run `nix flake lock`/`check`/`metadata` in this environment, so everything below is from reading `flake.lock` as JSON and walking the resolved input graph, plus the local `credit-card-offer-adder` checkout.

### 1. Nothing pre-existing was re-pinned

I walked the resolved graph from `root` in both `git show 77fe56f:flake.lock` and the new lock, following `inputs` edges and resolving `follows` list-values against `root`, then compared the locked identity (`type`/`owner`/`repo`/`rev`/`ref`/`url`/`narHash`/`lastModified`/`path`) at every reachable path:

- paths in both whose locked identity changed: **none**
- paths removed: **none**
- paths added: 11 (all under `credit-card-offer-adder/`)

The naive per-node-name diff does show seven nodes "changing" (`nixpkgs_2`–`nixpkgs_6`, `flake-utils_3`, `systems_2`) and every removed line in the raw diff is a `_N` key or edge target. That is exactly the renumbering shift you'd expect from inserting nodes into an alphabetically-keyed map — no node name was dropped, and the graph comparison above confirms none of it is semantic.

### 2. The added subtree is correct and minimal

Eight new nodes, which is precisely the expected closure and nothing more:

`credit-card-offer-adder`, `nixpkgs_2`, `flake-utils_2`, `systems_3`, `bun2nix`, `flake-parts`, `systems_2`, `treefmt-nix`

This matches the offer-adder's own `flake.nix` (`nixpkgs` = `nixos-unstable`, `flake-utils`, `bun2nix`, `bun2nix.inputs.nixpkgs.follows = "nixpkgs"`) edge for edge.

**Exactly one extra nixpkgs**, as intended. All transitive consumers resolve to the same `nixpkgs_2`:

- `credit-card-offer-adder.inputs.nixpkgs` → `nixpkgs_2`
- `bun2nix.inputs.nixpkgs` → `["credit-card-offer-adder", "nixpkgs"]` → `nixpkgs_2`
- `flake-parts.inputs.nixpkgs-lib` → `["credit-card-offer-adder", "bun2nix", "nixpkgs"]` → `nixpkgs_2`
- `treefmt-nix.inputs.nixpkgs` → same → `nixpkgs_2`

So `bun2nix` does follow the offer-adder's nixpkgs rather than dragging in a third one, and the deliberate non-`follows` on the root input is what the lock actually reflects: root `nixpkgs` stays at `d407951` and `nixpkgs_2` is an independent `9fbb54b`.

Note `systems_2` is `nix-systems/triplet` (bun2nix), not a duplicate of the `nix-systems/default` nodes — `systems_3` is the `default` one under `flake-utils_2`.

### 3. The pinned rev is the current tip of the offer-adder's main

- `rev` = `110fd306cf663bf5072407b345ffdb280171790a`, which in the local checkout is `origin/main` exactly (not an ancestor, not an off-branch commit) — "Merge pull request #2 from cprussin/claude/credit-card-offers-automation-k723hw".
- `lastModified` = `1788025654` = 2026-08-29 17:47:34 UTC, which matches that commit's timestamp (2026-08-29 10:47:34 -0700) to the second. Not stale.
- `type`/`owner`/`repo` are `github`/`cprussin`/`credit-card-offer-adder`, `original` carries no `ref` (default branch), and `narHash` is present. `narHash` is present on all eight new nodes.

### 4. No structural problems

- `version` is `7`, unchanged; `root` is `root`.
- Every `inputs` edge (including every `follows` list) resolves to a node that exists — no dangling references.
- Every node is reachable from `root` — no orphans.
- File is valid JSON with a trailing newline.
- No `flake: false` node was added; the six existing ones are untouched.
- On dedup: `flake-utils_2` and `systems_3` are byte-identical to the pre-existing `flake-utils`/`systems` nodes, and `nixpkgs_3`/`nixpkgs_5` were already identical before this change. That's inherent to the lock format — Nix does not dedupe across subtrees without explicit `follows` in `flake.nix` — and adding a `flake-utils.follows` here would work against the deliberate isolation of this input. Correct as-is.

### 5. The `nixpkgs.follows` rationale — consistent, but not fully verifiable here

`docs/DEPLOYMENT.md` in the offer-adder repo states the constraint the `flake.nix` comment summarises: Playwright refuses a browser bundle whose revision doesn't match the library, so `package.json`'s catalog pins `playwright` to a bare `1.61.1` (no caret), meant to equal `pkgs.playwright-driver.version`. The structure is consistent with the comment — the input resolves to its own `nixpkgs_2` (`9fbb54b`, 2026-08-26) rather than root's `d407951` (2026-08-04), so the pairing is decided by the offer-adder's own nixpkgs as the comment claims.

**I could not confirm that `9fbb54b` actually ships `playwright-driver` 1.61.1** — that needs `nix eval nixpkgs#playwright-driver.version` against that rev, which I can't fetch here. Flagging it as unverified rather than guessing; the structural claim holds either way, and if the versions were mismatched it would be a defect in the offer-adder repo, not in this lock.

One thing worth knowing for the future: the offer-adder repo has **no committed `flake.lock`** (only `bun.lock` is tracked, and `flake.lock` isn't gitignored). So its transitive pins live solely in this lock file, which is fine for this consumer, but it means a future `nix flake update credit-card-offer-adder` here floats `nixpkgs_2` forward, and the playwright/chromium pairing only stays valid if the catalog pin moves in lockstep — which is exactly the workflow `DEPLOYMENT.md` documents. No action needed for this PR.

### Out of scope, pre-existing — not introduced here

`flake.nix` declares `nixpkgs.url = "github:NixOS/nixpkgs/nixos-26.05"`, but the lock's `nixpkgs` node records `original.ref = "nixos-unstable"`. That mismatch is identical at `77fe56f`, at `1c6c06c`, and on `origin/main`, so this PR neither introduces nor fixes it. It does mean the next full re-lock will re-point root `nixpkgs` from unstable to 26.05 — a large, unrelated change that someone will hit eventually. Mentioning it only so it isn't a surprise later; nothing to do in this PR.

---

cprussin:
Two things from the round-4 review chased down, one of which corrects a premise in the PR description.

**The ntfy version I tested against was wrong.** The description said 2.26.0 — that is what nixpkgs *26.05* ships, but the lock's root `nixpkgs` node records `original.ref = "nixos-unstable"` (pre-existing, unchanged by this PR), and the rev it actually pins, `d407951`, ships **ntfy 2.25.0**. So every claim in the description about the auth provisioning was verified against a version that will not run on crux. `eval crux` passing does not catch this: the provisioning goes through freeform env vars, so a version without it would evaluate clean and fail at runtime.

Re-ran the whole matrix on 2.25.0, with the same `server.yml` this PR generates and an environment file from the real key command:

| request | result | want |
|---|---|---|
| anonymous publish to `offers-alerts` | 403 | 403 |
| `Bearer` token publish to `offers-alerts` | 200 | 200 |
| `Bearer` token publish to an off-ACL topic | 403 | 403 |
| human basic-auth publish to `offers-codes` | 200 | 200 |
| human basic-auth, wrong password | 401 | 401 |
| `POST /v1/account/token` as a provisioned user | 200 | 200 |

And end to end through the app's own `ntfyCodeSource` against 2.25.0: published the alert with the provisioned token, streamed the reply topic, picked up `902531` published back by the "phone" user. Identical to 2.26.0 — config provisioning landed in 2.11.0, so both are well past it. Description corrected to say 2.25.0, with 2.26.0 noted as also verified.

**The playwright pin you flagged as unverifiable checks out.** `raw.githubusercontent.com` serves by rev even though the API does not, so: at `9fbb54b` — the nixpkgs this lock pins for `credit-card-offer-adder` — `pkgs/development/web/playwright/driver.nix` is `version = "1.61.1"`, and the offer-adder's `package.json` catalog pins `"playwright": "1.61.1"` bare. Exact match, which is the condition `docs/DEPLOYMENT.md` says must hold or the first `launchPersistentContext` fails. That was the one remaining thing that could have broken the deployment at runtime with everything green.

Also confirmed at the locked root rev: the `ntfy-sh` module has `environmentFile`, and `jq` is 1.8.1 (the object-value syntax is parenthesised now, so it would work on 1.7 too).

Nothing to change in the diff.

---

PR #28, opened 2026-08-29
[crux] Add a Chromium build environment

cprussin:
Makes `crux` able to build Chromium, for the `domicile` engine-fork spike. New
`config/machines/crux/chromium-build.nix`, imported from `crux`'s
`default.nix`, plus a one-line `options` passthrough in `lib/zfs.nix`.

## What it does

- **`/build`** — `tank-fast/chromium`, a dedicated dataset. Machine-scoped
  regenerable scratch rather than user state, which is why it isn't a `/home`
  dataset: those are persisted and backed up, and on crux one would land on
  `tank` with all the entanglement described under *The pool choice* below.
- **`nofail` + `x-systemd.before=local-fs.target`** — `nofail` so a missing
  dataset can't fail `local-fs.target` and leave crux in `emergency.target`
  with no `sshd`, on the box that serves the house's DNS. `nofail` also drops
  the mount's implicit `Before=local-fs.target` (`systemd.mount(5)`), which the
  tmpfiles rules depend on, so `x-systemd.before=` restores the ordering
  without making the mount required again. `mkZfsFileSystems` gains an
  `options` passthrough for this, defaulting to exactly what it emitted before.
- **`setup-build-mount`** — `x-systemd.before=` fixes the *boot* ordering but
  not the *deploy* one: `switch-to-configuration` restarts
  `sysinit-reactivation.target` (re-running `systemd-tmpfiles`) and blocks on
  it before starting new mounts, so on the switch that introduces `/build` the
  rules land on the tmpfs root and the mount hides them. This unit redoes them
  after the mount, and fails — degrading the system — when `/build` isn't
  `tank-fast/chromium`.
- **`TMPDIR=/build/tmp`** — `/tmp` is tmpfs, so it is RAM; a Chromium link
  there would take the machine out. `sessionVariables`, not `variables`: the
  primary user's shell is nushell, which sources neither `/etc/profile` nor
  home-manager's `hm-session-vars.sh`. Aged `10d`, since moving off tmpfs loses
  the reboot-empties-it property.
- **`programs.nix-ld`** — the toolchain gclient hooks fetch is prebuilt,
  dynamically-linked binaries that can't run against the store unaided.
- `curl`, `git`, `lsb-release` (depot_tools probes it), `python3`.

## Verified

`alejandra --check`, `statix check`, `deadnix --fail` pass; CI green on the
previous head including **`eval crux`** and **`eval lyra`** — the latter is what
[...]

PR #29, opened 2026-08-30
[crux] Document how Chromium actually gets built

cprussin:
Comment-only follow-up to [#28](https://github.com/cprussin/dotfiles/pull/28),
recording what the first real build turned up. No behaviour change — the diff
is entirely comments in `config/machines/crux/chromium-build.nix`.

## Why

The build worked, but getting there cost several dead ends the module said
nothing about, and each one looked like a different problem than it was.

- **It goes through upstream's `tools/nix` shell, not a bare `autoninja`.**
  `gn gen` fails on a missing `pkg-config` otherwise — a missing *host tool*,
  which reads like the nix-ld case in the existing comment but isn't, since
  nix-ld only helps prebuilt binaries find libraries.
- **The `path:` prefix on the flake ref is load-bearing.** Without it the ref
  resolves to the enclosing git repo (`git+file:///…?dir=tools/nix`) and nix
  copies Chromium's **tracked tree** into the store first. Gitignored paths are
  excluded — and `out/` is gitignored, i.e. most of the 97 GB — which is the
  only reason that's slow rather than catastrophic.
- **`--command` silently does nothing, and `NIX_SHELL_RUN` is how to script
  it.** Not an argument-parsing problem: nix passes everything after that flag
  through untouched. Upstream's shell is a `buildFHSEnv` whose `shellHook`
  execs bwrap before the `exec` line nix appends is ever reached, so you land
  in an interactive bash having run nothing — which then looks like every
  subsequent command mysteriously failing, because they're going to a different
  shell than you think. `NIX_SHELL_RUN='…' nix develop "path:./tools/nix"` is
  upstream's documented escape hatch, and it works with `nix develop` as well
  as `nix-shell`.
- **`TMPDIR` is nested, not preserved.** `nix develop` drops the derivation's
  value and then exports a fresh `mktemp -d -t nix-shell.XXXXXX`, created under
  the caller's — so the shell sees `/build/tmp/nix-shell.XXXXXX`. bwrap leaves
  `/build` visible at the same path, so the build's temp files still land on
  the dataset rather than tmpfs, which is the point. Those per-shell
  directories are never removed on exit, which is what the `10d` tmpfiles age
  from #28 quietly handles.

## The measured cost

This is what the `domicile` side is deciding against, so it belongs in the repo
rather than only in a PR thread:

| | |
|---|---|
| Wall clock | **4h16m21s** — first clean build, cold |
| CPU time | ~60.7 core-hours (57.5 user + 3.2 sys) |
| Effective parallelism | ~14.2× across 16 jobs (~89% efficiency) |
| Disk | **97 GB** of the 200 GB quota (**48.5%**) |
| Steps | 56,376 executed (55,962 local + 414 no-exec), plus 29,340 reported skipped |

GN args were `is_debug=false symbol_level=0 is_component_build=true` with the
Wayland/Ozone flags — the cheap end of the range. Symbols or a second `out/`
configuration would move the disk figure substantially against that 48.5%,
which is the number to watch rather than the wall clock.

## `programs.nix-ld`

I said in #28 that if upstream's shell proved sufficient, nix-ld should be
dropped rather than carried for nothing. Half of that is now settled and half
isn't, and the comment says which is which.

**It cannot have been used during the build.** `buildFHSEnv` synthesises its own
`/lib64` inside the bwrap namespace, holding a real loader, so nix-ld's stub is
never bound in — that follows from what the shell is, without needing an
experiment.

**Whether anything else needs it is untested.** The case that would matter is
the first `fetch chromium`, which is necessarily driven from outside that shell
since there's no checkout yet. The comment names the one-line test — turn the
option off, switch, and run a fetched binary directly — rather than my earlier
claim that settling it costs a full rebuild. It doesn't; nothing about the
question involves compiling anything.

## Verified

`alejandra --check`, `statix check`, `deadnix --fail` pass, and the module still
evaluates against nixpkgs 26.05. Comments only, so no config change to check
beyond that.

Four review rounds went into the accuracy of these comments. Rounds 1 and 2
each found mechanisms asserted confidently for behaviour that had only been
observed by its result — including one where round 1's correction was itself
wrong and round 2 restored the original. The claims above are the ones that
survived independent verification: `nix develop` experiments, nixpkgs 26.05
source, and upstream's own `tools/nix`.

### DIFF
--- a/config/machines/crux/chromium-build.nix
+++ b/config/machines/crux/chromium-build.nix
@@ -1,3 +1,32 @@
+# Building Chromium here means going through upstream's own Nix shell, which is
+# what supplies pkg-config and the other host tools gn probes for:
+#
+#     cd /build/chromium/src
+#     nix develop "path:./tools/nix"
+#
+# The path: prefix matters -- without it the flake ref resolves to the
+# enclosing git repo and nix copies Chromium's tracked tree into the store
+# first.  Gitignored paths are excluded, which is the only reason that is
+# merely slow rather than catastrophic: out/ is most of the checkout.
+#
+# --command does not work here, and not for the reason it looks like: nix
+# passes everything after that flag through untouched, but upstream's shell
+# is a buildFHSEnv whose shellHook execs bwrap before nix's appended `exec
+# <command>` is ever reached, so you land in an interactive bash having run
+# nothing.  Upstream ships an escape hatch for scripting it:
+#
+#     NIX_SHELL_RUN='autoninja -C out/Domicile chrome' nix develop "path:./tools/nix"
+#
+# TMPDIR is nested rather than preserved: nix develop drops the derivation's
+# value, then exports a fresh mktemp directory created under the caller's, so
+# the shell sees /build/tmp/nix-shell.XXXXXX.  bwrap leaves /build visible at
+# the same path, so the build's temp files still land on the dataset rather
+# than tmpfs, which is the point.  Those per-shell directories are not removed
+# on exit -- that is what the age on the tmpfiles rule below is for.
+#
+# Measured 2026-08-29, first clean build, is_component_build with
+# symbol_level=0: 4h16m wall on 16 jobs, 97G of the 200G quota.  Both would
+# grow substantially with symbols or a second out/ configuration.
 {
   config,
   pkgs,
@@ -92,6 +121,14 @@ in {
   # gclient hooks, which cannot run against the store unaided.  If one dies on
   # a missing .so, add it to programs.nix-ld.libraries -- systemPackages will
   # not help.
+  #
+  # It is not consulted inside upstream's shell: that is a buildFHSEnv, so the
+  # loader at /lib64 in the container is the FHS env's own rather than nix-ld's
+  # stub.  Whether anything still needs it is untested -- the first `fetch
+  # chromium` is necessarily driven from outside that shell, so that is where
+  # it would earn its place.  Testing costs a switch and a second rather than a
+  # rebuild: turn this off and run a fetched binary directly, e.g.
+  # third_party/llvm-build/Release+Asserts/bin/clang --version.
   programs.nix-ld.enable = true;
 
   environment = {


config/machines/crux/chromium-build.nix
```diff
@@ -1,3 +1,19 @@
+# Building Chromium here means going through upstream's own Nix shell, which is
+# what supplies pkg-config and the other host tools gn probes for:
+#
+#     cd /build/chromium/src
+#     nix develop "path:./tools/nix"
+#
+# The path: prefix matters -- without it the flake ref resolves to the
+# enclosing git repo and nix copies Chromium's whole tracked tree into the
```
cprussin: Verified — this one is right, and worth keeping.

Reproduced with nix 2.35.2 on a throwaway git repo with a flake in `tools/nix`:

```
$ nix flake metadata ./tools/nix --json | jq -r .resolvedUrl
git+file:///.../outer?dir=tools/nix
$ nix flake metadata "path:./tools/nix" --json | jq -r .resolvedUrl
path:/.../outer/tools/nix
```

and forcing `self` to materialise in the first case produced a store path containing the whole repo (`bigfile.bin` and all), not just `tools/nix`.

One small precision, if you want it: what gets copied is the git tree — tracked files at HEAD when the checkout is clean, and the working tree minus `.gitignore` matches when it's dirty (I confirmed a gitignored `out/` is excluded). For a `gclient`-synced Chromium checkout that's the second case, so "whole tracked tree" is close but not literally what happens. Not worth a rewrite unless you're editing this block anyway.


---

config/machines/crux/chromium-build.nix
```diff
@@ -1,3 +1,19 @@
...
+# store first.  Enter the shell rather than passing --command: nix eats the
+# inner `-c` and drops you in an interactive bash having run nothing.  TMPDIR
```
cprussin: This mechanism is wrong. The symptom you saw is real, the advice is right, but the cause isn't nix's argument parser — and the real cause has a documented workaround you'd probably want.

**Nix does not eat the inner `-c`.** `--command`'s handler in `src/nix/develop.cc` takes a `std::vector<std::string>`, which gives the flag `ArityAny` (`args.hh`: the `std::function<void(std::vector<std::string>)>` constructor sets `arity(ArityAny)`), and `Args::processFlag` then does `for (n < ArityAny) { args.push_back(*pos++) }` — it consumes every remaining argument verbatim, with no further flag parsing. Identical code in 2.18.1, 2.24.9, 2.28.3 and 2.35.2.

Empirically, with nix 2.35.2:

```
$ nix develop <ref> --command bash -Z foo     # -Z is not a nix flag
$ nix develop <ref> --command bash -c "echo hi"
$ nix develop <ref> --command bash
```

all three produce byte-identical output (just the flake error for a bogus `<ref>`). If the parser were re-reading flags after `--command`, the first would have died on `unrecognised flag '-Z'`. It doesn't; `-Z`, and `-c`, land in the command vector.

**What actually happens.** Upstream's shell is `(pkgs.buildFHSEnv { ... }).env` (`tools/nix/make-shell-for-system.nix`). In nixpkgs, `buildFHSEnv` is `buildFHSEnvBubblewrap`, and its `.env` is a derivation whose `shellHook` is the bwrap invocation, ending in `exec "${cmd[@]}"`. `nix develop` builds an rc file that ends with `eval "${shellHook:-}"` and *then* appends `exec <your command>`. The shellHook execs bwrap, so the appended line is never reached. Inside the container `runScript` runs, and chromium's is:

```sh
if [ -n "${NIX_SHELL_RUN:+set}" ]; then eval "$NIX_SHELL_RUN"; else exec bash; fi
```

`NIX_SHELL_RUN` is unset, so: interactive bash, having run nothing. Exactly what you saw, and it's a property of `buildFHSEnv` shells generally, not of this flag.

**The useful bit you're missing.** Upstream documents both the failure and the fix. `tools/nix/README.md`:

> To run a command in this env, you can't use `nix-shell --run`, but instead:
> ```sh
> $ NIX_SHELL_RUN='...' nix-shell
> ```

and the `runScript` carries the comment "workaround for `nix-shell --run '...'` not working with buildFHSEnv". Same applies to `nix develop -c`. So the build *can* be driven non-interactively — `NIX_SHELL_RUN='autoninja -C out/Default chrome' nix develop "path:./tools/nix"` — which matters a lot more for a 4-hour build than "enter the shell manually".

Suggested replacement for the two sentences:

```
# The path: prefix matters -- without it the flake ref resolves to the
# enclosing git repo and nix copies Chromium's whole tracked tree into the
# store first.  --command does not work: upstream's shell is a buildFHSEnv,
# whose shellHook execs bwrap before nix's appended `exec <command>` line is
# reached, so you land in an interactive bash having run nothing.  To script
# it, use upstream's documented escape hatch instead:
#
#     NIX_SHELL_RUN='autoninja -C out/Default chrome' nix develop "path:./tools/nix"
```

There *is* a genuine `--command` arg-order trap worth a clause if you want one, just not this one: because the flag is `ArityAny`, everything after it is swallowed, so a flake ref or an `--impure` placed after `--command` silently becomes part of the command.


---

config/machines/crux/chromium-build.nix
```diff
@@ -92,6 +108,11 @@ in {
   # gclient hooks, which cannot run against the store unaided.  If one dies on
   # a missing .so, add it to programs.nix-ld.libraries -- systemPackages will
   # not help.
+  #
+  # Whether this is still needed is unverified: the first successful build ran
+  # entirely inside upstream's Nix shell, which supplies libraries itself, so
+  # nix-ld may never have been consulted.  Establishing that either way costs a
+  # full rebuild outside that shell.
```
cprussin: The hedge is the right instinct, but both halves can be sharpened without spending four hours.

**"may never have been consulted" is settled — it wasn't.** Upstream's shell is `pkgs.buildFHSEnv`, i.e. a bubblewrap namespace with a synthesized FHS root: `buildFHSEnv.nix` does `ln -s /usr/lib64 $out/lib64` and the container binds that over `/lib64`. nix-ld works by `environment.ldso`, which tmpfiles-links `/lib64/ld-linux-x86-64.so.2` to `${nix-ld}/libexec/nix-ld` on the host. Inside the container that path is the FHS env's own glibc loader, not nix-ld's stub. So nothing run inside upstream's shell can have used nix-ld — that's structural, no experiment needed. "Supplies libraries itself" undersells it: it supplies the loader too.

**"costs a full rebuild" isn't true.** Nothing about the question involves compiling anything. It's one `exec` of one prebuilt binary, outside the shell:

```
$ readelf -l third_party/llvm-build/Release+Asserts/bin/clang | grep interpreter
      [Requesting program interpreter: /lib64/ld-linux-x86-64.so.2]
$ third_party/llvm-build/Release+Asserts/bin/clang --version
```

If the interpreter is the FHS path, that binary can only run outside the container because something provides `/lib64/ld-linux-x86-64.so.2` — which on this machine is nix-ld and nothing else. Flipping `programs.nix-ld.enable = false` and re-running that one command is a `nixos-rebuild switch` plus a second, not 4h. (Same check applies to `buildtools/linux64/gn` and `third_party/rust-toolchain/bin/rustc`; Go-built tools like `cipd` are static and prove nothing.)

**And the interesting case is the fetch, not the build.** `fetch chromium` / `gclient sync` necessarily runs before `tools/nix` exists in the checkout, i.e. outside the shell — which is exactly the scenario the original comment above describes ("prebuilt binaries fetched by gclient hooks"). If nix-ld earned its place, it earned it there. That's also where a future reader will hit it again, so it's worth saying.

I'd rather see this block say what is known and what the cheap test is than say "unverified, costs 4h" — the latter reads as "don't bother", which is how an option ends up carried forever. Something like:

```
# nix-ld is not consulted inside upstream's shell: that shell is a
# buildFHSEnv, so /lib64's loader inside the container is the FHS env's own,
# not nix-ld's stub.  Whether anything still needs it -- gclient sync runs
# outside that shell -- is untested.  Testing is cheap, not a rebuild: flip
# this off, switch, and run a fetched binary directly, e.g.
# `third_party/llvm-build/Release+Asserts/bin/clang --version`.
```


---

config/machines/crux/chromium-build.nix
```diff
@@ -1,3 +1,19 @@
...
+#
+#     cd /build/chromium/src
+#     nix develop "path:./tools/nix"
+#
+# The path: prefix matters -- without it the flake ref resolves to the
+# enclosing git repo and nix copies Chromium's whole tracked tree into the
+# store first.  Enter the shell rather than passing --command: nix eats the
+# inner `-c` and drops you in an interactive bash having run nothing.  TMPDIR
+# survives into that shell (nix nests its own temp dir inside it), so the
+# checkout's temp files stay off tmpfs as intended.
```
cprussin: Verified. `TMPDIR` is in `Common::ignoreVars` in `src/nix/develop.cc` (alongside `TMP`, `TEMP`, `TEMPDIR`), so the value from the derivation's environment is dropped rather than exported, and the caller's survives. The parenthetical is right too — `nix develop` does `createTempDir("", "nix-develop")`, which lands under whatever `TMPDIR` is.

Worth knowing that the bwrap layer doesn't undo it either: `buildFHSEnv`'s `privateTmp` defaults to false, and top-level host directories that aren't part of the FHS root get auto-mounted, so `/build` is visible inside the container at the same path.


---

cprussin (commented):
Comment-only diff, so this is purely an accuracy review. I checked each claim against nix's own source (2.18.1 / 2.24.9 / 2.28.3 / 2.35.2), by experiment with nix 2.35.2 locally, and against upstream's actual `tools/nix` (readable from the chromium GitHub mirror). Three of the four claims hold up; one is wrong, and the nix-ld hedge understates what is already known while overstating what it would cost to settle. Details inline.

## Verified correct

- **The shell supplies `pkg-config`.** Upstream's `tools/nix/make-shell-for-system.nix` lists `pkg-config` in `targetPkgs`, annotated with the exact `gn gen` failure you'd hit without it (`FileNotFoundError: [Errno 2] No such file or directory: 'pkg-config'`). The whole `targetPkgs` list is annotated that way, incidentally — it's a good reference for the next missing-library surprise.
- **The `path:` prefix is load-bearing**, for the stated reason. Reproduced.
- **`TMPDIR` survives.** `TMPDIR` is in `nix develop`'s `ignoreVars`, so the derivation's value is dropped and the caller's kept.

## Wrong

- **The `--command` mechanism.** Nix does not eat the inner `-c` — `--command` is `ArityAny` and swallows the rest of the line verbatim, in every nix version I checked. The real cause is that upstream's shell is a `buildFHSEnv` whose `shellHook` execs bwrap before nix's appended `exec <command>` is ever reached. Upstream documents this and ships a workaround (`NIX_SHELL_RUN='...'`) that lets you drive the build non-interactively — which seems worth having for a 4-hour job. Inline.

## Overstated

- **"Establishing that either way costs a full rebuild."** It costs one `exec` of one prebuilt binary. And half the question is already settled structurally: a `buildFHSEnv` container supplies its own `/lib64` loader, so nix-ld cannot have been consulted inside that shell. Inline.

## Numbers

Everything derived in the PR body checks out:

| Claim | Check |
|---|---|
| 4h16m21s wall | `real 256m20.986s` = 4h16m20.99s |
| 57.5 user + 3.2 sys | 3450m32.994s = 57.509h; 193m31.699s = 3.226h |
| ~60.7 core-hours | 60.734 |
| ~14.2x parallelism | 3644.08 CPU-min / 256.35 wall-min = 14.215 |
| ~89% efficiency | 14.215 / 16 = 88.8% |

The committed comment's "4h16m wall on 16 jobs, 97G of the 200G quota" agrees with the body. Two nits, both body-only:

- 97/200 is 48.5%, written as 48% twice. Worth rounding up or writing ~48.5%, given you're pointing at that number as the one to watch.
- The step breakdown doesn't visibly add up: 56,376 total against 55,962 local + 0 remote leaves 414 unaccounted, and 29,340 skipped on a *first clean* build wants a word. Either say what those are or keep only the total.

## Style

Header comment above the argument attrset is fine — `config/modules/data/session-vars/home.nix` already does the same thing, so there's precedent. No other machine module has a file header, but none of them has anything to say before the attrset either.

---

config/machines/crux/chromium-build.nix
cprussin: You're right, and I've rewritten it. Reproduced your experiment before changing anything, since this was my own confident explanation of something I'd watched happen:

```
$ nix develop "path:/nonexistent-xyz" --command bash -Z foo
$ nix develop "path:/nonexistent-xyz" --command bash -c "echo hi"
$ nix develop "path:/nonexistent-xyz" --command bash
```

All three, byte-identical — just the flake error. `-Z` is never rejected, so nothing after `--command` is being re-parsed as a flag.

There's independent corroboration for your `buildFHSEnv` explanation too: the handoff document that started this work already used `NIX_SHELL_RUN='...' nix-shell tools/nix/shell.nix`. I read that as a stylistic choice and reached for `--command` instead, then explained the resulting breakage with a mechanism I'd invented rather than checking why upstream documents that form.

The comment now gives the real cause and the escape hatch. `NIX_SHELL_RUN` is the more valuable half — "enter the shell manually" is fine advice for a person, useless for anything scripted, and this is a four-hour job.

Left out the `ArityAny` arg-order trap. It's real, but it's a fact about nix rather than about building Chromium on this host, and this comment is already long.

---

config/machines/crux/chromium-build.nix
cprussin: Taken, close to your wording. Both halves of my hedge were wrong in the same direction — I treated an unknown as unknowable.

The structural half is settled and I should have reasoned it out rather than shrugging: `buildFHSEnv` synthesises its own `/lib64`, so nix-ld's stub is shadowed inside the container and nothing run in that shell could have reached it. That's derivable from what the shell *is*, no experiment needed.

The cost half was worse. "Costs a full rebuild" was me anchoring on the 4h number I'd just measured, when the question involves compiling nothing — it's `readelf` plus one `exec`. And you're right about how that lands: "unverified, costs 4h" is functionally "leave it alone forever", which is exactly how an unexplained option outlives the thing it was added for.

The comment now says what's known, names `gclient sync` as where it would actually earn its place, and gives the one-line test. Whoever picks this up gets a decision they can make in a minute instead of a warning not to try.

---

config/machines/crux/chromium-build.nix
cprussin: Took the precision, since I was editing the block anyway. It now says nix "copies the whole checkout into the store" rather than "whole tracked tree" — accurate for the dirty-working-tree case a gclient-synced checkout is actually in, and not wrong for the clean one either.

Thanks for reproducing it with `flake metadata`; `git+file://...?dir=tools/nix` versus `path:/.../tools/nix` is a much sharper demonstration than "it seemed to hang", which is all I had.

---

config/machines/crux/chromium-build.nix
cprussin: Thanks — `ignoreVars` is the actual mechanism, and my parenthetical had the causality backwards. I'd written that nix "nests its own temp dir inside" `TMPDIR`, which is a consequence rather than the reason; the reason is that nix deliberately drops the derivation's `TMPDIR` and keeps the caller's.

The `privateTmp = false` point is the one that closes it, and I hadn't thought about the bwrap layer at all. Two independent things had to hold for `/build/tmp` to survive into the build — nix not overriding it, and the container not hiding it — and I'd only observed the composite result. Both are now in the comment.

---

cprussin:
Second review round found three more inaccuracies; all fixed in `9865107`. Recording them here since the reviewer hit GitHub's rate limit and couldn't post inline.

**`TMPDIR` — wrong again, and my round-1 "fix" made it worse.** I originally wrote that nix "nests its own temp dir inside" `TMPDIR`. Round 1 told me the mechanism was `ignoreVars` keeping the caller's value, and I took it. That's wrong: nix drops the derivation's value *and then* exports a fresh `mktemp -d -t nix-shell.XXXXXX`, which lands under the caller's `TMPDIR`. Measured directly rather than taking either account on faith:

```
$ TMPDIR=/tmp/mytmp nix develop path:… --command bash -c 'echo $TMPDIR; echo $NIX_BUILD_TOP'
/tmp/mytmp/nix-shell.Wdg3mF
/tmp/mytmp/nix-shell.Wdg3mF
```

Which is exactly the `/build/tmp/nix-shell.3dA5Th` we saw on crux. My original wording was closer than the correction I accepted, and I should have tested it the first time instead of deferring — the conclusion held either way, so nothing surfaced the error. The comment now also notes that those per-shell directories are never removed on exit, which is what the `10d` tmpfiles age quietly handles.

**"the whole checkout" overstates the `path:` hazard.** The git fetcher copies the tracked tree; gitignored paths are excluded — and `out/` is gitignored, which is most of the 97 GB. Round 1's "whole tracked tree" was the better phrasing and I loosened it. Now says tracked tree, and says why that's the only thing keeping the hazard to "slow" rather than "catastrophic".

**"gclient sync runs before tools/nix exists" is not true.** `tools/nix` is part of `src`, so it's there as soon as `src` is cloned. What's actually true is narrower: the *first* `fetch chromium` has to be driven from outside the shell, since there's no checkout yet. Reworded.

Two rounds running, the errors have been in mechanisms I asserted confidently for behaviour I'd only observed the result of. The conclusions have all survived; the explanations haven't.

Not changing `out/Domicile` — the reviewer noted it appears nowhere else in the repo and upstream's convention is `out/Default`. It's the project's own output directory name from #28, and nothing in this repo creates it either way.

---

cprussin:
Third review round: **the comment content is accurate, no errors in the code.** Every mechanism was re-verified independently — `nix develop` experiments against nix 2.35.2, nixpkgs 26.05's `build-fhsenv-bubblewrap`, and upstream's real `tools/nix` from the chromium mirror. Posting the outcome here because the reviewer hit GitHub's rate limit again and couldn't submit inline.

What it confirmed, beyond the three claims this revision rewrote:

- The rc script `nix develop` generates literally ends `export NIX_BUILD_TOP="$(mktemp -d -t nix-shell.XXXXXX)"` / `export TMPDIR="$NIX_BUILD_TOP"`, with the derivation's value never restored. `-t` is what parents it under the caller's.
- `--command` passes its arguments through untouched — `nix develop -c sh -c 'echo "$0 | $*"' --offline --impure "a b"` prints `--offline | --impure a b`, nothing mangled. The failure really is `eval "${shellHook:-}"` preceding `exec '<command>'` in that same rc.
- `NIX_SHELL_RUN` works under `nix develop`, not just `nix-shell`, because nix develop doesn't clear the caller's env and `bwrapCmd` never passes `--clearenv`.
- `/build` survives into the container because `bwrapCmd` binds every top-level directory the FHS env doesn't itself provide.
- nix-ld shadowing confirmed against a built `*-fhsenv-rootfs`: `lib64 -> /usr/lib64` holds a real loader and lands in `ignored`, so the host `/lib64` stub is never bound in.

Three findings, all addressed in `d9cccd7`:

1. **The PR body contradicted the file.** The substantive one — it still carried the round-1 story (`whole checkout`, `TMPDIR` "keeps the caller's value", `gclient sync`, a `readelf` step the comment doesn't use). The commit message was right; only the body was stale. It's now rewritten, which matters because the body is what a reader lands on from blame while the file is the durable artifact, and the two disagreed on precisely the mechanism that took three rounds to pin down.
2. **Line 129 ran to 93 columns** against 80 elsewhere. Rewrapped.
3. **"nix swallows everything after that flag"** could be read as "nix eats the arguments" — the exact wrong diagnosis that sentence exists to head off. Now "passes everything after that flag through untouched".

One thing the review noted that I've left alone: nix leaves *two* directories per shell, not one (`nix-shell.XXXXXX` and `nix-develop-<pid>-<rand>`). The comment says "those per-shell directories", which covers it.

---

PR #57, opened 2026-09-13
crux: a DRM card the Domicile tty work can actually test on

cprussin:
Implements the three things the crux handoff asked for, in one new file
(`config/machines/crux/domicile-tty.nix`) rather than folded into
`domicile-ci.nix` or `claude-agent.nix` — those are each about a *way of
working* on the engine, and this is the hardware underneath them, which
either one uses and neither one owns.

## What is in it

- **`video` group for the primary user.** This affects *ssh sessions only*,
  which is narrower than the handoff assumed. Two things could already open
  the card: the CI runner, because `domicile-ci.nix` gives its unit
  `SupplementaryGroups = ["render" "video"]` and systemd resolves that group
  *name* through NSS without consulting its member list; and a session at
  the physical console, through logind's uaccess ACL (see limit 2). What
  this adds is interactive ssh and `claude-agent`.
- **`vkms`**, with `enable_cursor=1 enable_writeback=1 enable_overlay=1`.
  All four connectors on the 970 read `disconnected` and no `video=`
  kernel argument fixes that (the proprietary driver ignores them), so a
  virtual connector is the only one this machine can have. The writeback
  connector is what turns "it did not crash" into a pixel verdict.
- **`drm-acl`**, a wrapper plus a NOPASSWD sudo rule, to flip which card
  node the user can open.

```
$ drm-acl status
card0        nvidia-drm   not hidden from cprussin
card1        vkms         not hidden from cprussin
renderD128   nvidia       not hidden from cprussin
$ sudo drm-acl hide nvidia-drm    # Chromium's primary is now vkms
$ sudo drm-acl show nvidia-drm    # back
```

("not hidden" rather than "open": `status` looks for the deny entry and
nothing else, so it does not speak to mode or group.)

## Two costs of item 1, both written into the file

The console is on that card, so anything running as that user can become
DRM master and blank it. crux is headless, so that costs nothing until the
day something has gone wrong enough to need a console.

The quieter one: `nvr.nix` sets `ffmpeg.hwaccel_args = "preset-nvidia"`, so
the house cameras decode on this same 970. A Chromium GPU process competes
with those NVDEC sessions, and *that* failure is dropped camera recording —
which nobody notices the way they notice a dead console.

## A wrapper, not `sudo setfacl`

Which the handoff said it would prefer. `setfacl` as root takes any path;
this validates against a closed set before building anything, and names one
user resolved at build time, so `/dev/dri` is the whole of its reach.

It takes **a driver name** (`nvidia-drm`, `vkms`) as well as a node name,
and the driver is the spelling to use: `cardN` numbering is registration
order, so naming `card0` names a slot, and the day that order changes,
`hide card0` hides the wrong card and the test runs against a blind GPU —
silently. Naming the driver cannot do that. It refuses rather than guesses
if one driver has two nodes, and a driver resolves to that driver's *card*
node only, which is what Chromium selects on. `status` reports the driver
behind each node for the same reason.

## Three limits, documented rather than engineered around

Each of these wants a policy decision this file shouldn't make alone. Happy
to do any of them as a follow-up.

1. **The ACL lasts one boot.** `/dev` is devtmpfs, so the nodes come back
   bare and nothing here reapplies them. The bad case is silent: crux
   reboots, the hide is gone, the next run tests the blind card and reports
   whatever a blind card reports. A run that cares should check `drm-acl
   status` rather than assume. Making it persist is a udev rule plus a
   marker file.
2. **logind can undo it inside a boot.** systemd tags `SUBSYSTEM=="drm",
   KERNEL=="card*"` for uaccess, so a login at the physical console
   rewrites the same named-user entry to `rw-` for that session, and drops
   it on logout rather than restoring the deny. Local seats only, so ssh is
   unaffected.
3. **The sudo half does not reach inside a CI job**, and the handoff
   assumed it would ("the runner runs as cprussin, so anything granted to
   that user covers both"). True for the `video` group; not for sudo. The
   github-runner module sets `NoNewPrivileges` (mkDefault true) and an
   empty `CapabilityBoundingSet` (mkBefore) — the first makes the kernel
   ignore the setuid bit, the second leaves a setuid-root exec holding no
   capabilities at all. I left both alone rather than relaxing a runner
   that already runs trusted-workflows-only as a wheel user. If CI should
   drive the switch itself, the choice is relaxing those two or going
   around them (a template unit plus a polkit rule escalates
   out-of-process, which neither directive touches).

## Not done, because the handoff said not to

No kernel command line change, no change to `nvidia-drm` (`hardware.nix`
sets `modesetting.enable`, and with nvidia ≥ 545 nixpkgs already emits both
`nvidia-drm.modeset=1` and `nvidia-drm.fbdev=1` — confirmed in
`nixos/modules/hardware/video/nvidia.nix`), and no X or Wayland session.

Also deliberately **no `fbcon=map:`**. vkms brings a second framebuffer, but
fbcon claims only the *first* one registered and moves later only where
`con2fb_map_boot[]` sends it — so vkms, loaded from `systemd-modules-load`
long after nvidia-drm has fb0, cannot take the console. That parameter is
the one thing that *could* move the console onto a display nobody can see,
so it is not added as a precaution.

## Testing

Now rebased onto #58, so **`eval crux` has evaluated this module for real
and passes**, along with `eval lyra`, `flake check` and the rest — all 17
jobs green. (Before that rebase, `eval crux` aborted on the insecure-immich
failure before reaching anything here, so this is its first genuine
evaluation.)

Locally, with a nix installed for the purpose:

- `alejandra --check .`, `statix check .`, `deadnix --fail .` — all clean.
- Evaluated through `eval-config.nix` against a stubbed `primary-user`:
  `extraGroups` → `["video"]`, `boot.kernelModules` → includes `vkms`,
  `extraModprobeConfig` → the expected `options vkms …` line, sudoers rule →
  `{users = ["cprussin"]; command = "/run/current-system/sw/bin/drm-acl";
  options = ["NOPASSWD"];}`.
- Built `drm-acl` and exercised it against a fake `/dev/dri` and
  `/sys/class/drm` in a mount namespace: driver-name selection resolves to
  the right node, an ambiguous driver is refused, and validation rejects
  `../../etc/shadow`, `card0;rm`, a bare `-`, a missing node, no arguments,
  and `status` with an argument.
- Verified the ACL semantics for real — char device, throwaway user in a
  `video`-like group: in the group → access; after `setfacl -m u:X:---` →
  denied; after `setfacl -x u:X` → access again. Also that `-x` on an
  absent entry exits 0, so `show` is idempotent, and that the `getfacl -cE
  | grep -qxF` detection reads both states correctly.

**What none of that covers:** nothing here has run against a real DRM
device. Whether `vkms` is actually built in this kernel, whether its
writeback connector produces the pixels the test wants, and whether
Chromium picks the card the ACL leaves open are all facts about the booted
machine.

So after the deploy and a reboot, the handoff's own checklist is what
settles it — `id cprussin | grep -o video`, `lsmod | grep vkms`, `ls
/dev/dri/` for a second card node, `sudo -n drm-acl status`. If `/dev/dri/`
still shows only `card0` and `renderD128`, `vkms` did not load; that is the
one most likely to go wrong quietly, and `modinfo vkms` is the check.

https://claude.ai/code/session_01BGHoEW5cyqdgaRMsP3kZhM


cprussin:
CI: `eval crux` is red, and it is not this PR's.

It fails on `immich-2.7.5` being newly marked insecure in nixpkgs (CVE-2026-59258, CVE-2026-82272, plus "2.x.x will not receive further updates"), which is nothing this diff touches — immich comes in through `photos.nix`. The same job is the sole failure on `main` at dd18c8e (["[flake] update sources"](https://github.com/cprussin/dotfiles/actions/runs/34773962579)), the commit this branch is cut from, so it arrived with the flake bump rather than with anything here. The other 16 jobs are green on both.

No re-run: this is deterministic, not a flake.

A separate agent is working the immich failure on its own branch now. Once that fix exists I will port it into this branch so this PR can go green without waiting for that one to merge.

---

PR #61, opened 2026-09-16
chore: add Renovate configuration

cprussin:
## What

Adds `renovate.json` at the repo root, based on `config:recommended` with:

- **`minimumReleaseAge: "7 days"`** — Renovate won't propose an update until a release has been out for 7 days. This is the whole point: it lets the bot do the dependency-bump legwork automatically instead of an agent doing it manually, while still giving a window for upstream regressions to surface before we pull them in.
- **Automerge policy** — minor/patch/pin/digest updates automerge; major updates are left for manual review (`packageRules`).
- **Coverage**:
  - **GitHub Actions** (`.github/workflows/ci.yml`) — covered out of the box by `config:recommended`, no extra config needed.
  - **Nix flake inputs** (`flake.nix`/`flake.lock`) — covered via the `nix` manager, which I had to enable explicitly (`"nix": {"enabled": true}`). It is **not** on by default under `config:recommended` — Renovate's own docs ship its default as `"enabled": false`. I also enabled `lockFileMaintenance` explicitly, since that's off for every manager by default too, and most of this repo's flake inputs (`nixpkgs`, `home-manager`, `nixos-hardware`, etc.) track branches rather than a pinned rev, so per-input PRs won't fire for them — only a lock file maintenance run will pick them up.

Validated with `renovate-config-validator renovate.json` — passes.

## Activation

**This config does nothing on its own.** Renovate only acts once the [Renovate GitHub App](https://github.com/apps/renovate) is installed on this repo (or the account) by someone with admin access. I can't install it myself — that's a manual step for a human.

---

https://claude.ai/code/session_01SUgm7gyGctSgRbofXMevWx

---
_Generated by [Claude Code](https://claude.ai/code/session_01SUgm7gyGctSgRbofXMevWx)_


renovate.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": ["config:recommended"],
+  "timezone": "UTC",
+  "minimumReleaseAge": "7 days",
```
cprussin, line 5: **[code-review, single-pass without Agent tool]** `minimumReleaseAge` gates PRs by `releaseTimestamp`, which git-commit/digest updates (the update type nix flake.lock bumps almost always use) typically do not have. So for exactly the dependency type this config enables (`nix.enabled: true`), the 7-day wait may not actually apply, and combined with `automerge: true` for `digest` updates, flake input bumps (including nixpkgs) can auto-merge immediately with no cooling-off period -- defeating the apparent intent of `minimumReleaseAge`.

---

renovate.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": ["config:recommended"],
+  "timezone": "UTC",
```
cprussin, line 4: **[code-review, single-pass without Agent tool]** `timezone` only affects `schedule`-based cron windows. Nothing in this config (nor, by default, `config:recommended`) sets a `schedule`, so this key currently has no effect -- dead config that could be dropped (or paired with an actual `schedule` if a specific window was intended).

---

renovate.json
```diff
@@ -0,0 +1,16 @@
...
+  "nix": {
+    "enabled": true
+  },
+  "lockFileMaintenance": {
+    "enabled": true
+  },
+  "packageRules": [
+    { "matchUpdateTypes": ["minor", "patch", "pin", "digest"], "automerge": true },
+    { "matchUpdateTypes": ["major"], "automerge": false }
+  ]
```
cprussin, line 20: **[code-review, single-pass without Agent tool]** `automerge: false` for `major` updates is redundant -- Renovate already defaults `automerge` to `false`, and there is no earlier rule or preset setting in this file that a `major` update would otherwise inherit `true` from. Could be dropped, unless it is meant purely as explicit documentation of intent.

---

renovate.json
cprussin: Good catch — fixed by excluding the `nix` manager from automerge entirely rather than relying on `minimumReleaseAge` for digest-based flake.lock updates (a5911d3).

---

renovate.json
cprussin: Fair — no `schedule` is set, so this is inert today. Leaving it in as explicit intent for if/when a schedule window is added; not worth a push on its own.

---

renovate.json
cprussin: Agreed, redundant with Renovate's default — kept as explicit documentation of intent alongside the minor/patch/pin/digest rule above it. Not worth a push on its own.

---

